PR #2, opened 2020-04-29
Create initial blueprints for deploying Kubeflow.

jlewi:
* This PR contains two blueprints

  * A management blueprint to setup a management cluster running KCC
    This can be used to create GCP resources

  * A kubeflow blueprint for deploying kubeflow.
    * This currently doesn't deploy all the Kubeflow applications
      because some of the Kubeflow applications need to be updated
      to kustomize v3 so they can be incorporated into the package.

* Related to kubeflow/manifests#1063

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,23 @@
+# Reset various kpt values to default values and remove other
+# files that shouldn't be included in PRs
+# TODO(jlewi): We should add a test to make sure changed values don't get checked in
+# We don't run it in generate because we don't want to force all developers to install kpt
+clean-for-pr:
+	rm -rf kubeflow/.build
+	rm -rf management/.build
+	
+	rm -rf kubeflow/upstream/manifests
+	rm -rf management/upstream/management
+	
+
+	kpt cfg set ./kubeflow/instance name KUBEFLOW-NAME
+	kpt cfg set ./kubeflow/instance gcloud.core.project PROJECT
+	kpt cfg set ./kubeflow/instance mgmt-ctxt MANAGEMENT-CTXT
+
+	kpt cfg set ./management/instance name MANAGEMENT-NAME
+	kpt cfg set ./management/instance gcloud.core.project HOST_PROJECT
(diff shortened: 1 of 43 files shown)


rmgogogo:
Hi Jeremy,

(1)
If I want to follow the existing code repo to deploy via KCC, which doc can help I quickly catch the steps? This PR is ready for an installation or not ready but it's the initial PR. If ready, I may just follow the doc to do a test which would help more.

this one?: management/README.md

(2)
Is this target to Onprem/Anthos, or GCP? We have several options to install KF/P. Would like to get your input on targeted use case / scenario.

jlewi:
@rmgogogo If you want to follow the instructions you need to do two things

1. Follow management/README.md to create a management cluster

    * A management cluster can be reused across projects and deployments and isn't specific to Kubeflow
1. Follow kubeflow/README.md to create a Kubeflow deployment

> Is this target to Onprem/Anthos, or GCP? We have several options to install KF/P. Would like to get your input on targeted use case / scenario.

Blueprints aren't really targeted to specific use case. They could be reused across Anthos and GCP with different modifications. For example, on Anthos you wouldn't create the GCP infra because that already has to exist.

The main thing about the blueprint is that its entirely declarative and using the new tools (KPT, KCC, AnthosCLI) being developed to support declarative management of GCP.

Bobgy:
Great work!
/lgtm

Some questions below

# {"type":"string","x-kustomize":{"setBy":"kpt","

Is this format written by kpt and can be understood by kustomize?

We have been using kustomize vars for parameters, this seems like a good alternative. 

How mature is kpt for production? It solves some of our problems with kubectl.

jlewi:
@Bobgy I think GCP has started recommending kpt for various pieces (e.g. setting up ASM) but its still early. I think its safe to adopt because the concern is likely just backwards incompatible changes or API changes. But that can be mitigated by just installing the old version of kpt and not updating.

/approve

Issue #5, opened 2020-05-04
Setup autodeploy for GCP blueprints


jlewi:
We should setup the auto-deploy infrastructure to autodeploy from blueprints.

This way we ensure that our GCP blueprint is up to date and working.

jlewi:
auto-deploy is running on cluster `kf-ci-v1`.

I made the service account `kf-ci-v1-user@kubeflow-ci.iam.gserviceaccount.com` an owner of projects in folder `ci-projects` so that it would be able to deploy in project kubeflow-ci-deployment.

jlewi:
This was working so I'm closing this issue.
It recently broke and kubeflow/testing#668 is tracking that.

PR #7, opened 2020-05-05
Fix various bugs as part of setting up auto deployments

jlewi:
* Fix some bugs in the blueprints that cropped up while working on
  setting up continuous auto-deployments using the blueprints (#5)

Fix some bugs in the documentation.

* Fix bugs in the management config for the per namespace components
  of CNRM. The namespaces of the role bindings wasn't correct so
  the cnrm manager pod ended up not having appropriate permissions.

  * Also the scoped namespace of the cnrm manager statefulset needs
    to be set the managed project not the host project.

* Update Makefile to point at kubeflow/manifests master to pull in cert-manager
changes.

* Add check_domain_length to validate the length of the hostname KF
  deployment name so that we don't end up exceeding the certificate limits.

### DIFF
--- a/kubeflow/Makefile
+++ b/kubeflow/Makefile
@@ -5,6 +5,8 @@
 MGMTCTXT=$(shell yq r ./instance/settings.yaml mgmt-ctxt)
 # The name of the context for your Kubeflow cluster
 NAME=$(shell yq r ./instance/settings.yaml name)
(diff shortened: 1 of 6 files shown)


jlewi:
/assign @rmgogogo 
/assign @Bobgy 

Could one of you review this please?

PR #17, opened 2020-05-19
Fix env var in README.md

Bobgy:
/assign @jlewi


Bobgy:
I'm not sure if I missed anything. I cannot run those commands and I need to change it instead.
Error message:
```
➜  management kpt cfg set ./upstream name $(MGMT_NAME)
zsh: command not found: MGMT_NAME
```

Issue #13, opened 2020-05-15
Management cluster should not use CNRM in namespaced mode


jlewi:
We are currently recommending installing CNRM in namespaced mode.

I'm not sure thats what we should recommend.

* Namespace mode is inconvenient when managing multiple projects as we end up having to create
  multiple deployments of the CNRM system

* Using ACM to install CNRM doesn't appear to install it in namespaced mode.

jlewi:
We should cherry pick the fixes onto the 1.1 branch and then we can close this.

PR #19, opened 2020-05-19
Fix cnrm instance commands

Bobgy:
/assign @jlewi

### DIFF
--- a/management/README.md
+++ b/management/README.md
@@ -111,12 +111,12 @@ For each project you want to setup follow the instructions below.
 1. Create a copy of the per namespace/project resources
 
    ```
-   cp -r ./instance/install-per-project ./instance/cnrm-install-${MANAGED_PROJECT}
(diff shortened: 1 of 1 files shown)


management/README.md
```diff
@@ -111,20 +111,20 @@ For each project you want to setup follow the instructions below.
 1. Create a copy of the per namespace/project resources
```
jlewi, line 111: Per: https://github.com/kubeflow/gcp-blueprints/issues/13 
I think we just want to install CNRM in non-namespace mode using workload identity. Would you mind changing this to something like  "follow the instructs https://cloud.google.com/config-connector/docs/how-to/install-upgrade-uninstall to install CNRM using Cloud Config Connector using workload identity in non namespace mode"

Bobgy: Got it, SGTM

jlewi: Are you planning on making any changes?

Bobgy: @jlewi Sorry, I've got some other urgent issues, won't have time to follow up on this one.

Issue #33, opened 2020-05-28
GCP Blueprint for KF on private GKE


jlewi:
Uber issue for creating a blueprint for installing Kubeflow on private GKE

jlewi:
I've run into a couple issues getting this to work

* kubeflow/kfctl#340 - The way kfctl mirror works today won't work with the directory structure
   used by blueprints so we will need to fix that.

* kubeflow/gcp-blueprints#36 - Cloud Endpoints Controller won't work with the restricted VIP APIS.

* Per kubeflow/gcp-blueprints#34 I tried setting up Cloud Nat and Firewall rules to allow pulling images from DockerHub and Quay.IO. 
    * I was able to make DockerHub work but not Quay.IO. 
    * I was hoping this might provide a simpler alternative to mirror'ing images but I don't think this would work with VPC-SC since the existing kubeflow-images-public registry would not be in the service perimeter

jlewi:
I have various PRs in progress to fix the mirror pipelines:

* kubeflow/kfctl#345 - kustomize function to transform images
* kubeflow/kfctl#342 - kfctl isn't recoursing over dependencies correctly figure out all images to be replicated

jlewi:
So I got this working; a couple follow on pieces:

kubeflow/kfctl#342  - Still need to get this reviewed and merged 
#44 - Updates the Makefile to use kfctl to apply the blueprint.

After the above are merged only remaining step would be to update the docs:
https://www.kubeflow.org/docs/gke/private-clusters/

jlewi:
Only remaining step is getting the docs updated.

jlewi:
Initial docs are in the markdown
https://github.com/kubeflow/gcp-blueprints/blob/master/kubeflow/deploy_private.md

PR #40, opened 2020-06-04
Additional instructions and fixes for private GKE

jlewi:
* Tracking issue #33

* The rules "apply-asm", "apply-gcp", "apply-kubeflow" should all
  depend on the hydrate rule; not component specific hydration rules.

  * This is because we might want to apply kpt functions as part of hydration
    so these commands would be added to the hydration rule and need
    to be called before calling the respective apply commands.

* Delete enable-services.yaml we no longer use it; we will use CNRM.

* A bunch of improvements to the Makefile

  * Organize it in hydrate and apply sections
  * Add a set-values stanza to run all the kpt cfg set commands
  * Use if statements to deal with differences e.g. private vs.
    non-private deployments
     * Need to add this for ASM deployments.

### DIFF
--- a/kubeflow/Makefile
+++ b/kubeflow/Makefile
@@ -6,6 +6,7 @@ MGMTCTXT=$(shell yq r ./instance/settings.yaml mgmt-ctxt)
 # The name of the context for your Kubeflow cluster
 NAME=$(shell yq r ./instance/settings.yaml name)
 PROJECT=$(shell yq r ./instance/settings.yaml project)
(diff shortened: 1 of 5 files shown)


rmgogogo:
LGTM. (didn't provide label so that Yuan can still take  a look).
Let us know if you prefer submit this PR quickly.

Bobgy:
/lgtm
forgot about this, thanks @rmgogogo for reminder

PR #41, opened 2020-06-06
Wait for cert-manager webhook to be created.

jlewi:
* We observe transient failures when deploying kubeflow with the failure

Error from server (InternalError): error when creating ".build/kubeflow-apps/cert-manager.io_v1alpha2_certificate_admission-webhook-cert.yaml": Internal error occurred: failed calling webhook "webhook.cert-manager.io": the server is currently unable to handle the request

* The problem appears to be that the cert-manager webhook isn't available yet

* Add a wait statement to wait for the webhook deployment to be available.


jlewi:
@rmgogogo going forward please use "/approve" if you're ok with the PR being merged. 

/approve

Issue #43, opened 2020-06-08
Kubeflow deployment fails "webhook.cert-manager.io" unavailable


jlewi:
Deploying Kubeflow fails the first time you run it with the error

```
Error from server (InternalError): error when creating ".build/kubeflow-apps/cert-manager.io_v1alpha2_certificate_admission-webhook-cert.yaml": Internal error occurred: failed calling webhook "webhook.cert-manager.io": the server is currently unable to handle the request
```

The problem is its trying to create the certificate for the Kubeflow admission controller and failing because certmanager isn't available yet.

Simply waiting and retrying fixes the problem. Is there a better solution though?

jlewi:
Waiting for the deployments to be ready is insufficient.

```
	kubectl --context=$(KFCTXT) -n cert-manager wait --for=condition=Available --timeout=600s deploy cert-manager-webhook
	kubectl --context=$(KFCTXT) -n cert-manager wait --for=condition=Available --timeout=600s deploy cert-manager
	kubectl --context=$(KFCTXT) -n cert-manager wait --for=condition=Available --timeout=600s deploy cert-manager-cainjector
```

It looks like the cert-manager may not have any readiness or health probes which could be the problem.

jlewi:
The upstream deployment appears to have probes
https://github.com/jetstack/cert-manager/blob/master/deploy/charts/cert-manager/templates/webhook-deployment.yaml#L68

Ours do not.
https://github.com/kubeflow/manifests/blob/master/cert-manager/cert-manager/base/deployment.yaml

jlewi:
It looks like this might have been the PR that added the liveness and readiness probes.
https://github.com/jetstack/cert-manager/pull/2245

It looks like the first cert-manager release for that was 0.12. It looks like we are currently on 0.11.

jlewi:
Filed: kubeflow/manifests#1234 it looks like we have to fix the manifest upstream. 
Short term fix would be to add retries to the blueprint. I'll leave this issue open to track

1. Adding retries to fix this
1. Removing retries once the upstream fix is available.

PR #44, opened 2020-06-09
Minor tweaks to the Makefile for the blueprint

jlewi:
Add more wait statements for cert-manager.
  * Related to: #43
  * Looks like this doesn't actually help the cert-manager problem; looks
    like we may need to update cert-manager so we can get health checks.

* Use kfctl to apply the cloud endpoints resource; the code for cloud-endpoints-controller should now be baked into kfctl.

### DIFF
--- a/kubeflow/Makefile
+++ b/kubeflow/Makefile
@@ -201,6 +201,8 @@ apply-kubeflow: hydrate
 	kubectl --context=$(KFCTXT) apply -f ./$(BUILD_DIR)/cert-manager
 	# We need to wait for certmanager webhook to be available other wise we will get failures
 	kubectl --context=$(KFCTXT) -n cert-manager wait --for=condition=Available --timeout=600s deploy cert-manager-webhook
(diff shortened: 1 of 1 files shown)


kubeflow/Makefile
```diff
@@ -201,6 +201,8 @@ apply-kubeflow: hydrate
 	kubectl --context=$(KFCTXT) apply -f ./$(BUILD_DIR)/cert-manager
 	# We need to wait for certmanager webhook to be available other wise we will get failures
 	kubectl --context=$(KFCTXT) -n cert-manager wait --for=condition=Available --timeout=600s deploy cert-manager-webhook
+	kubectl --context=$(KFCTXT) -n cert-manager wait --for=condition=Available --timeout=600s deploy cert-manager
```
Bobgy, line 204: why duplicate?
Shall we do
```
kubectl --context=$(KFCTXT) -n cert-manager wait --for=condition=Available --timeout=1800s deploy cert-manager
```

Bobgy: ohh, never mind, they are different deployments

PR #45, opened 2020-06-09
Add kpt setters for default profile.

jlewi:
* We need kpt setters in order to properly set the email for the default
  profile created.

* Also rerun `clean-for-pr` to clear some of the default values.


Bobgy:
/approve
/lgtm

Right, I had to manually fix some of these when setting up my own cluster

Issue #55, opened 2020-06-17
GCP Blueprint autodeployments failing - problems with pipelines


jlewi:
GCP autodeployments; IAP isn't enabled.

Looks like the `kubeflow-oauth` secret is missing.

jlewi:
It looks like the problem is the deployment pipelines are failing.

Looking at the logs the problem is

```
INFO|2020-06-18T16:34:57|/workspace/testing-repo/py/kubeflow/testing/util.py|72| The PersistentVolumeClaim "minio-pvc" is invalid: spec: Forbidden: is immutable after creation except resources.requests for bound claims
INFO|2020-06-18T16:34:57|/workspace/testing-repo/py/kubeflow/testing/util.py|72| Makefile:191: recipe for target 'apply-kubeflow' failed
INFO|2020-06-18T16:34:57|/workspace/testing-repo/py/kubeflow/testing/util.py|72| make: *** [apply-kubeflow] Error 1
WARNING|2020-06-18T16:34:58|/workspace/testing-repo/py/kubeflow/testing/create_kf_from_gcp_blueprint.py|198| make apply failed with retryable error. The output matched regex: .*webhook\.cert-manager\.io.*
INFO|2020-06-18T16:34:58|/workspace/testing-repo/py/kubeflow/testing/util.py|46| Running: make apply 
cwd=/workspace/blueprint-repo/kubeflow
INFO|2020-06-18T16:34:58|/workspace/testing-repo/py/kubeflow/testing/util.py|56| Running: Environment:
```

Looking higher in the logs it looks like we already created the PVC.

```
INFO|2020-06-18T16:34:24|/workspace/testing-repo/py/kubeflow/testing/util.py|72| persistentvolumeclaim/minio-pvc created
INFO|2020-06-18T16:34:24|/workspace/testing-repo/py/kubeflow/testing/util.py|72| secret/mlpipeline-minio-artifact created
INFO|2020-06-18T16:34:25|/workspace/testing-repo/py/kubeflow/testing/util.py|72| service/minio-service created
```

@Bobgy My guess is this is an issue with recent changes related to the pipeline manifest.

My guess is you are duplicating the PVC resource.

Bobgy:
Yes, that's right. I think we still need a gcp stack  in kubeflow/manifests to compose resources without duplication.

I'll send a PR for that

Bobgy:
@jlewi Can you take a look at the two PRs above as a whole?
what do you think?

[...]

PR #57, opened 2020-06-19
Use gcp stack for blueprint

Bobgy:
/assign @jlewi 
Part of https://github.com/kubeflow/gcp-blueprints/issues/55

Depends on https://github.com/kubeflow/manifests/pull/1286 being merged first


Bobgy:
Thanks
/unhold

PR #58, opened 2020-06-22
fix: add pipeline mysql install config override

Bobgy:
/assign @jlewi 
Sorry, this is the fix to override the configmap from instance

Fix https://github.com/kubeflow/manifests/issues/1294


jlewi:
Thanks.

So it looks like we use CNRM to create the disks.
https://github.com/kubeflow/manifests/blob/master/gcp/v2/cnrm/pipelines/disk.yaml

And then use overlays to change the name of the disk to a name set via the setter.

Bobgy:
Yes, that's right

PR #62, opened 2020-06-23
Set user email to default profile

Bobgy:
/assign @jlewi 

default profile doesn't get an email address, setting it in this PR.


Bobgy:
note, there are quite a bit of auto formatting

sounds good we just clean them up?

Issue #74, opened 2020-07-09
knative and kfserving is missing from GCP Blueprint


jlewi:
test knative is deployed is failing.
https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/kubeflow-gcp-blueprints-periodic-master/1281020489434664960

jlewi:
This on the master branch of kubeflow/manifests. I want to verify it is working on the master auto deployment before cherry-picking to the 1.1. branch.

Once we verify on master we will need to do the following

1. Cherry-pick it into kubeflow/manifests 1.1 branch
1. Create a kubeflow/manifests rc
1. Update the blueprint to the new RC.

jlewi:
Test is failing against master.

Looking at one of the auto-deployed clusters the knative namespace is empty.

```
kubectl -n knative-serving get deploy
No resources found in knative-serving namespace.
```

jlewi:
#1376 had a bug; it would end up installing knative in the kubeflow namespace which isn't what we want.

jlewi:
We will need to update the blueprint Makefile to install knative

jlewi:
knative deployment is failing on master.

```
Error from server (NotFound): error when creating ".build/knative/~g_v1_configmap_config-tracing.yaml": namespaces "knative-serving" not found
```

It looks like the knative package includes the namespace resource

```
.build/knative/~g_v1_namespace_knative-serving.yaml 
```

But we need to apply it in order

jlewi:
The test is still failing but it looks like knative is now installed on master

```
kubectl -n knative-serving get deploy
NAME               READY   UP-TO-DATE   AVAILABLE   AGE
activator          1/1     1            1           5h12m
autoscaler         1/1     1            1           5h12m
autoscaler-hpa     1/1     1            1           5h12m
controller         1/1     1            1           5h12m
networking-istio   1/1     1            1           5h12m
webhook            1/1     1            1           5h12m
```

Downgrading to P1 since this looks like an issue with the test.

PR #87, opened 2020-07-14
Need to create the knative-serving namespace in the proper order

jlewi:
Related to #74


jlewi:
/assign @Bobgy 

@ellistarn Would you mind LGTM'ing this? @Bobgy is in Shanghai ; probably won't be around for another 6-8 hours?

ellistarn:
The syntax is throwing me
```
knative/~g_v1_namespace_knative-serving.yaml 
```
but I'll assume you know what you're doing. LGTM to add a knative namespace to install those resources in.
Additionally, we've had discussions about installing other components (i.e. KFServing) into their own namespaces. It may become slightly more challenging to maintain, but this is how the knative project does its namespaces, i.e. knative-serving and knative-eventing

jlewi:
@ellistarn Thanks. The filenames are auto-generated by kustomize I think the name scheme for the files is something like "api <group>.<namespace>.<resource name>"

Issue #89, opened 2020-07-15
Master auto-deployment is failing


jlewi:
Here are the logs

```
INFO|2020-07-15T04:02:18|/workspace/testing-repo/py/kubeflow/testing/util.py|72| set 3 fields
INFO|2020-07-15T04:02:19|/workspace/testing-repo/py/kubeflow/testing/util.py|46| Running: kpt cfg set ./upstream/manifests/gcp location us-central1-c 
cwd=/workspace/blueprint-repo/kubeflow
INFO|2020-07-15T04:02:19|/workspace/testing-repo/py/kubeflow/testing/util.py|61| Subprocess output:

INFO|2020-07-15T04:02:20|/workspace/testing-repo/py/kubeflow/testing/util.py|72| set 0 fields
INFO|2020-07-15T04:02:21|/workspace/testing-repo/py/kubeflow/testing/util.py|46| Running: kpt cfg set ./upstream/manifests/gcp email kf-ci-v1-user@kubeflow-ci.iam.gserviceaccount.com 
cwd=/workspace/blueprint-repo/kubeflow
INFO|2020-07-15T04:02:21|/workspace/testing-repo/py/kubeflow/testing/util.py|61| Subprocess output:

INFO|2020-07-15T04:02:22|/workspace/testing-repo/py/kubeflow/testing/util.py|72| set 0 fields
INFO|2020-07-15T04:02:23|/workspace/testing-repo/py/kubeflow/testing/util.py|46| Running: kpt cfg set ./upstream/manifests/stacks/gcp name kf-vbp-0715-a29 
cwd=/workspace/blueprint-repo/kubeflow
INFO|2020-07-15T04:02:23|/workspace/testing-repo/py/kubeflow/testing/util.py|61| Subprocess output:

INFO|2020-07-15T04:02:24|/workspace/testing-repo/py/kubeflow/testing/util.py|72| set 4 fields
INFO|2020-07-15T04:02:25|/workspace/testing-repo/py/kubeflow/testing/util.py|46| Running: kpt cfg set ./upstream/manifests/stacks/gcp gcloud.core.project kubeflow-ci-deployment 
cwd=/workspace/blueprint-repo/kubeflow
INFO|2020-07-15T04:02:25|/workspace/testing-repo/py/kubeflow/testing/util.py|61| Subprocess output:

INFO|2020-07-15T04:02:25|/workspace/testing-repo/py/kubeflow/testing/util.py|72| set 0 fields
[...]

PR #95, opened 2020-07-20
Convert setters from v1 to v2

jlewi:
* Get rid of enable-services.yaml as we are now using CNRM and this is
  in the base package.
* Related to #89

* Remove cluster and iam policy patch; these are now in the base manifests.
  Added in kubeflow/manifests#1398

* kubeflow/hack/create_kptfile.py contains most of the commands
  used to generate the setters and substitutions.

* Add a presubmit to verify the deploy pipeline is working.


jlewi:
/assign @Ark-kun 

@Ark-kun Could you review it since its very late in shanghai?

tests/pipeline-runs/deploy-blueprint-pipeline-run.yaml
```diff
@@ -0,0 +1,45 @@
...
+      params:
+        - name: revision
+          value: master
+        - name: url
+          value: https://github.com/kubeflow/gcp-blueprints.git
+  - name: testing-repo
+    resourceSpec:
+      type: git
+
+      # TODO(https://github.com/kubeflow/testing/pull/740): Change back
```
Bobgy, line 30: nit kubeflow/testing#740 already merged

kubeflow/hack/create_kptfile.py
```diff
@@ -0,0 +1,114 @@
+"""This is a script created for the updating of the GCP Kpt packages.
+
+The purpose of this is to upgrade to using setters and substitutions
```
Bobgy, line 3: nit: shall we add a comment to mention it should be in sync with kubeflow/manifest/hack/create_kptfile.py

jlewi: I think we can delete this in a subsequent PR; I just want to check it in so we have a record in source control but its only intended for one time use.

Bobgy: SGTM

kubeflow/instance/kustomize/kubeflow-apps/service-accounts.yaml
```diff
@@ -3,4 +3,4 @@ kind: ServiceAccount
 metadata:
   name: kf-admin
   annotations:
-    iam.gke.io/gcp-service-account: kf-kcc-0415-admin@PROJECT.iam.gserviceaccount.com # {"type":"string","x-kustomize":{"setBy":"kpt","partialSetters":[{"name":"name","value":"kf-kcc-0415-001"},{"name":"gcloud.core.project","value":"PROJECT"}]}}
+    iam.gke.io/gcp-service-account: kf-kcc-0415-admin@PROJECT.iam.gserviceaccount.com
```
Bobgy, line 6: @jlewi just noticed this is probably a mistake, there is no setter here

Issue #29, opened 2020-05-21
[ACM] Support structured repository


jlewi:
Currently our ACM story doesn't use a structured repository. The main reason was because
we would need to refactor the manifests; e.g. put ClusterScoped resources in one directory and namespace resources in a different directory.

Could we reorganize the files using a kustomize function?

Using a flat repository will get pretty unmanageable. 

Can we use subdirectories with an unstructured repositories?

jlewi:
Fixed by #109

Issue #14, opened 2020-05-15
ACM and iap enabler will conflict - switch to using RCToken


jlewi:
The IAP enabler pod will try to patch the policy with the backend service id which is the audience.  This has been a source of problems because the backend can change.

With ASM we should be able to use a configurable audience (RCTOKEN).
https://cloud.google.com/service-mesh/docs/iap-integration#configure_the_iap_access_list

This should allow us to pick and use a deterministic audience so we won't run into issues with the iap-enabler pod fighting ACM to update the ingress policy.

It doesn't look like we can set the RCToken on IAP using a GKEBackendConfig so we might need to  write a simple daemon to do that.

jlewi:
The IAP enabler pod is also configuring the health check.
https://github.com/kubeflow/manifests/blob/b3f5b41780a01845619884e638f4f6c9183d8fef/gcp/iap-ingress/base/config-map.yaml#L185

So we need to keep that bit even though we don't want to patch the audience

jlewi:
It looks like with RC Tokens a different header is used.
https://cloud.google.com/service-mesh/docs/iap-integration#enable_rctoken_support_on_the_service_mesh

jlewi:
The IAP docs are now referring to ISTIO RequestAuthentication resources which are introduced in ISTIO 1.5.
https://istio.io/latest/news/releases/1.5.x/announcing-1.5/upgrade-notes/#authentication-policy

I think we are currently on ISTIO 1.4.

jlewi:
I think I misspoke above regarding the health check. It is a separate statefulset 
https://github.com/kubeflow/manifests/tree/b3f5b41780a01845619884e638f4f6c9183d8fef/gcp/iap-ingress/base

That is updating the health check.  There is now a better approach for correctly setting the health check: 
kubeflow/gcp-blueprints#115

PR #109, opened 2020-08-02
Update instructions for using ACM.

jlewi:
* Use a kpt function to remove namespace from non namespace scoped
  objects

* Use yq to attach backend config to the ingress.

* Remove the iap enabler pod; this is a partial work around for #14

  * The IAP enabler pod will try to update the ISTIO security policy
    which will conflict with ACM. So we disable it for now even though
    that means we have to manually update the health check.

* Switch to using a structured repo with ACM (#29)

  * Add a script to rewrite the YAML files in the appropriate structure

  * If we don't use a structured repository we end up with problems because
    resources in different namespaces but with the same name will be written
    to the same file.

* Add a hack to create the kube-system namespace as part of the ACM deployment.

  * Now that we are using structured repositories we need to have
    a namespace directory with a namespace.yaml for kube-system
    in order to install resources in that namespace.

Related to #4 - use ACM to deploy Kubeflow


Bobgy:
/lgtm
/approve
ACM is definitely something I want to try when setting up test KFP with KFP. It's great we can support this use-case.

Issue #110, opened 2020-08-04
Auto deployments are failing - problem curling the meshconfig endpoint


jlewi:
```
curl --request POST \
--header "Authorization: Bearer `gcloud auth print-access-token`" \
--data '' \
--fail \
"https://meshconfig.googleapis.com/v1alpha1/projects/kubeflow-ci-deployment:initialize"
% Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
curl: (22) The requested URL returned error: 409
Makefile:189: recipe for target 'apply-asm' failed
make: *** [apply-asm] Error 22
```

It looks like this was added in #99 

@Bobgy Should we revert #99 from the 1.1 branch until we can figure out what the problem is?

Bobgy:
I have no idea why it got 409, can it be lack of permission?

According to the documentation, it should be a one time setup per project. So shall we move that step to documentation instead?

jlewi:
@Bobgy yes; moving to documentation sounds good at least until we figure it out.

PR #114, opened 2020-08-05
Revert "Cherry pick of "fix: ASM sidecar repeated auth error message to stackdriver #96""

Bobgy:
Reverts kubeflow/gcp-blueprints#99

Because of https://github.com/kubeflow/gcp-blueprints/issues/110


Bobgy:
@jlewi can you take a look at the presubmit failure

PR #119, opened 2020-08-12
fix: fix a few problems when deploying e2e

Bobgy:
Changes:
1. set-values problem
   `make set-values` fail with
   ```
   set-values:
   make: set-values:: No such file or directory
   make: *** [set-values] Error 1
   ```
1. remove outdated comment

### DIFF
--- a/kubeflow/Makefile
+++ b/kubeflow/Makefile
@@ -58,12 +58,10 @@ set-values:
 # The rules below help fetch and update packages
 #************************************************************************************************************************
 # Get packages
+# TODO(jlewi): We should think about how we layout packages in kubeflow/manifests so
+# users don't end up pulling tests or other things they don't need.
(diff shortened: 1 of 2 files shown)


kubeflow/Makefile
```diff
@@ -58,12 +58,10 @@ set-values:
 # The rules below help fetch and update packages
 #************************************************************************************************************************
 # Get packages
+# TODO(jlewi): We should think about how we layout packages in kubeflow/manifests so
+# users don't end up pulling tests or other things they don't need.
 .PHONY: get-pkg
 get-pkg:
-	# TODO(jlewi): We should switch to using upstream kubeflow/manifests and pin
```
Bobgy, line 63: These comments are printed when running make commands, so prefer moving them out

PR #120, opened 2020-08-12
Update management makefile

subodh101:
What?
- Define Name, Project and Zone_or_Region at one place.
- Remove extra `set-values` from the same rule.

Why?
- Creates confusion of editing the same values at two places.
- bug fix.


subodh101:
Let me know if there is no problem I will update the [docs](https://github.com/kubeflow/website/blob/master/content/en/docs/gke/deploy/management-setup.md) of the same.

subodh101:
@googlebot I signed it!

Bobgy:
I think this is a good idea.
What do you think? @jlewi 
/lgtm

Bobgy:
Another idea I had was that we can save these config in a JSON or yaml config file and let makefile read from there.
Because editing a makefile is at the end of day still not be best user interface to my opinion.

management/Makefile
```diff
@@ -1,33 +1,27 @@
...
-#***********************************************************************************************************************
-# Edit this section to set the values specific to your deployment
-
 .PHONY: set-values
 set-values:
-	set-values:	
-	kpt cfg set ./instance name <NAME>
-	kpt cfg set ./instance location <ZONE_OR_REGION>
-	kpt cfg set ./instance gcloud.core.project <PROJECT_TO_DEPLOY_IN>
+	kpt cfg set ./instance name $(NAME)
```
jlewi, line 18: Not using make variables was intentional.  The intent was to make the KptFile the source of truth. Hence the set-values stanza (i.e. calling kpt cfg) is the preferred way of setting the values.

I think if we wanted to dedupe the stanzas we could refactor it so we only set the values in one kpt package and then read it from that kpt package.

subodh101: Even if we refactor the kpt packages into one kpt package there needs to be a way to edit these values. I think it won't be changed.

The main concern at the moment is while deploying management cluster I was thinking like maybe `instance pkg` is for a kubeflow cluster and `upstream/management pkg` is for mgmt cluster. And with this minimal change, at least this won’t happen again with the next person trying to deploy the mgmt cluster.

subodh101:
With the new change step significantly improves also preserving the source of truth values.
```zsh
NAME=<NAME> PROJECT=<PROJECT> ZONE_OR_REGION=<ZONE_OR_REGION> make set-values
```
cc: @jlewi

jlewi:
/hold

Supporting the syntax

```
NAME=<NAME> LOCATION=<ZONE_OR_REGION> PROJECT=<PROJECT> make set-values
```

Means we are creating two ways of setting values; e.g. with Make and with kpt. I think we want kpt to be the source of truth.

If a user wants to set the value name or LOCATION the way they should do it is

```
kpt cfg set . name mykubeflow
```

The set-values rule is primarily sugar to reduce copy and paste of the kpt commands.

Instead of creating an extra layer ontop of `kpt cfg set` can we reduce the layers?

Specifically

1. How could we avoid duplicating the parameters when we have multiple kpt packages all of which share one or more variables?
1. When we upgrade we do `rm -rf upstream/management` so how can we preserve the values set by the user?

@subodh101 any thoughts on how we could addres these problems?

subodh101:
Kpt is still the only source of truth, not makefile. But we know that atleast a user has to set NAME, LOCATION and PROJECT once in the kpt. And the way to set values in the kpt is achieved by the above mentioned syntax.

Note that it only run once and the values get defined in the `instance/setting.yaml` which is used for multiple kpt packages.

Since the source of truth values get set in the `instance/setting.yaml` there will be no effect with the command `rm -rf upstream/management`. So the values will be preserved.

New change: Because I used environment variable. I have removed further use of variables in the makefile and getting values directly from the `instance/setting.yaml`.

jlewi:
@subodh101 this doesn't address my concern. I don't think we should be telling users to set values by passing variables to make; e.g. the following goal

```
NAME=<NAME> LOCATION=<ZONE_OR_REGION> PROJECT=<PROJECT> make set-values
```

is an antipattern IMO because this command is teaching people that to set NAME, PROJECT etc... they can override environment variables. Which is not what we want. The message should be they should run `kpt set...`. For example the above would lead people to expect that they can do

```
NAME=<name> make apply
``` 

and have NAME override name. 


My preference would be to move towards getting rid of the set-values stanza all together and tell folks to run

```
kpt cfg set . name <NAME>
kpt cfg set . location <location>
kpt cfg set . gcloud.core.project <project>
make apply
```

Hence my questions above about deduping the `kpt set` commands. 

Can we simply move the `kpt set  upstream` commands into hydrate? The values would then be set based on the values obtained from settings.yaml.

subodh101:
@jlewi 
I made sure that the environment variables won't override.

About the antipattern, I didn't think a user can do such a thing. LGTM. 

Getting rid of `set-values`. ✅

subodh101:
@jlewi 
Let's merge this if there is nothing else to do.
Also, I have updated the [docs](https://github.com/kubeflow/website/pull/2126/files) for the same. Let me know if that LGTY.

Bobgy:
Agree with what comes out from the discussion!
Teaching users to use kpt while reducing the one extra call seems good.

management/Makefile
```diff
@@ -1,33 +1,16 @@
-
-# The kname of the context for the management cluster
-# The name of the context for your Kubeflow cluster
 NAME=$(shell yq r ./instance/settings.yaml name)
+PROJECT=$(shell yq r ./instance/settings.yaml project)
```
Bobgy, line 2: Can we add an error message here if ./instance/settings.yaml is still empty/the default value?
similarly for others

subodh101: Since the setting values is one step also overriding values doesn't make any change with `kpt cfg set`. I have combined the values and made one make rule.

Output if values are not set.
```zsh
▶ make apply
either of <NAME>, <ZONE_OR_REGION>, <PROJECT_TO_DEPLOY_IN> not set.
make: *** [validate-values] Error 1
```

jlewi:
This is great thanks. Two additional suggestions

1. Can we get rid of settings.yaml and just get the values from the kpt file e.g
    https://github.com/kubeflow/gcp-blueprints/blob/2f797a19f24d3c6449bf311039cb0a5d38f5b563/kubeflow/instance/Kptfile#L21

1. nit: Can we keep the placeholders as "PROJECT", "LOCATION", etc... and not "dummy" ?

jlewi:
Thanks!

/lgtm

I'll let @Bobgy approve this so he can have one final look.

management/Makefile
```diff
@@ -46,20 +37,23 @@ apply: hydrate
...
 	rm -rf .build
 	mkdir -p .build/
 	mkdir -p .build/cluster
 	kustomize build $(INSTANCE_DIR)/cluster -o .build/cluster
 
-
-# Create a kubeconfig context for the kubeflow cluster
+# Create a kubeconfig context for the management cluster
 .PHONE: create-ctxt
 create-ctxt:
```
Bobgy, line 54: also depend on validate values here?

management/Makefile
```diff
@@ -1,38 +1,29 @@
-
-# The kname of the context for the management cluster
-# The name of the context for your Kubeflow cluster
-NAME=$(shell yq r ./instance/settings.yaml name)
+NAME=$(shell yq r ./instance/Kptfile 'openAPI.definitions."io.k8s.cli.setters.name".x-k8s-cli.setter.value')
```
Bobgy, line 1: nit: add a simple comment above, explaining Kptfile's values are set when calling kpt cfg set?

I think a goal of the makefile is to allow user customization. So it needs to be easy to understand.

subodh101: Right!

subodh101: Side Note: I ran the `kpt cfg set` with the default values to keep the values uniform throughout.

subodh101:
I looked into the [build logs](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubeflow_gcp-blueprints/120/kubeflow-gcp-blueprints-presubmit/1294633981626552321/#1:build-log.txt%3A813).

Questions regarding test failure:
- Is there any way to check Workflow `deploy-kf-master-g7x7p` pod logs? if not what would be the series of steps taking place in the pod? any script ref.(I could not find.)
- It seems Workflow running in the namespace `auto-deploy` is not present in previous PR's.

Bobgy:
/retest
let me take a look at the error log

Bobgy:
Hi @subodh101 I'm trying to fix presubmit test in https://github.com/kubeflow/gcp-blueprints/issues/128.

Bobgy:
Looks like presubmit tests are back normal

subodh101:
@Bobgy 
Thanks! Let's merge if everything else LGTY.

Bobgy:
/lgtm
/approve

I'd prefer we hold the documentation PR for a while, we can roll both out when we release 1.1.1

PR #133, opened 2020-09-10
Remove deprecated --project flag from anthoscli

plang85:
Flag --project has been deprecated, GCE project flag is unused and deprecated


Bobgy:
Thank you for the fix!
/LGTM
/approve

Issue #132, opened 2020-09-04
unable to recognize "./.build/istio/Base/Base.yaml": Get https://x.x.x.x/api?timeout=32s


Bobgy:
get the following error message when `apply-asm` step in kubeflow/Makefile
```
kubectl --context=dsp-kubeflow apply --recursive=true -f ./.build/istio/Base/Base.yaml
unable to recognize "./.build/istio/Base/Base.yaml": Get https://x.x.x.x/api?timeout=32s: dial tcp x.x.x.x:443: i/
```

reported in https://kubeflow.slack.com/archives/C7REE0EHK/p1599122036065700

The error message just means kubectl failed to reach the kubernete api server, but the most common way for getting this problem is that:
1. deploy a kubeflow instance
2. sth fails, so delete the kubeflow instance
3. fix some issues and redeploy a kubeflow instance with the same KF_NAME

Now, there's already a kubectl context with the same cluster name, so https://github.com/kubeflow/gcp-blueprints/blob/395902ee5f8069f3d0ed5c50a51f9aee62ce0c40/kubeflow/hack/create_context.sh#L19-L22
The create-ctxt.sh script won't create a new context instead, and the kubectl context will point to the deleted cluster deployed in step 1.

Bobgy:
the workaround is `kubectl config delete-context $KF_NAME`, so that when you run `make apply` again, it will create the new context for you.

Bobgy:
/cc @jlewi 
I'm thinking maybe we can just always refetch the kubectl context, it doesn't hurt in other ways, but it solves this purpose.

jlewi:
The error looks like the IP address is wrong. I'm guessing the recreated the cluster and the IP address changed?
Does it really make sense to recreate the context on every call? That seems like an unnecessary call to gcloud container clusters get-credentials which adds a bit of latency.

Maybe we could just document the error in the troubleshooting guide and tell people to recreate the context?

Bobgy:
Fixed by https://github.com/kubeflow/gcp-blueprints/pull/134.

PR #134, opened 2020-09-12
Update config context

subodh101:
What?
Get new config context `access-token` every time when trying to deploy/update kubeflow.

Why?
Config `access-token` gets changed if the cluster was deleted during operations which happen sometimes and it's very difficult for a person to debug that. As the resources, while applying gives a timeout error.
This brings a clean swipe install.
Related Issue: https://github.com/kubeflow/gcp-blueprints/issues/132


kubeflow/hack/create_context.sh
```diff
@@ -6,24 +6,12 @@
 # PROJECT=<PROJECT> REGION=<region> NAME=<NAME>
 #
 # TODO(jlewi): Support zonal clusters as well
-set -x 
+set -ex
 
 # Default namespace to kubeflow
 NAMESPACE=${NAMESPACE:-kubeflow}
-echo Checking if context ${NAME} exists
```
Bobgy, line 13: I believe you need to delete the context if it already exists

Bobgy (commented):
Thanks!

kubeflow/hack/create_context.sh
subodh101: Oh yes!
I was hoping to find an option like `--override` or `--force` in the rename context command. But unfortunately could not find. So using the previous conditional procedure again for checking the context.

kubeflow/hack/create_context.sh
```diff
@@ -6,24 +6,17 @@
...
-kubectl config use-context ${NAME}
-
-RESULT=$?
-
-if [ ${RESULT} -eq 0 ]; then
-echo kubeconfig context ${NAME} already exists
-exit 0
-fi
+# Delete kubeconfig context if exist already
+kubectl config delete-context ${NAME}
```
Bobgy, line 12: Recommend opt out one command explicitly sort of like
```
kubectl config delete-context ${NAME} || echo "OK - context ${NAME} does not exist"
```
It's also important to let people know if there is an error message there, it's expected. This helps when people debug other problems

subodh101: Updated error message based on the exact output
```zsh
error: cannot delete context ${NAME}, not in path/to/config
```

kubeflow/hack/create_context.sh
```diff
@@ -6,24 +6,17 @@
...
-kubectl config use-context ${NAME}
-
-RESULT=$?
-
-if [ ${RESULT} -eq 0 ]; then
-echo kubeconfig context ${NAME} already exists
-exit 0
-fi
+# Delete kubeconfig context if exist already
+kubectl config delete-context ${NAME} || echo "Ignore error - cannot delete context ${NAME}"
```
Bobgy, line 12: sounds good

nit: What do you think about? or do you have any other ideas?
`Context ${NAME} doesn't exist, so we will create a new one.`

The key information is why the context doesn't exist isn't a problem.

kubeflow/hack/create_context.sh
```diff
@@ -6,24 +6,17 @@
...
-
-kubectl config use-context ${NAME}
-
-RESULT=$?
-
-if [ ${RESULT} -eq 0 ]; then
-echo kubeconfig context ${NAME} already exists
-exit 0
-fi
+# Delete kubeconfig context if exist already
```
Bobgy, line 11: What do you think?
nit:
```
# Always delete the existing kubeconfig, because it may be an outdated
# context to a deleted cluster with the same name.
```

Bobgy (commented):
/assign @jlewi 
Can you take a look too?

/lgtm

kubeflow/hack/create_context.sh
subodh101: Does a user really need to get the details of what is going to happen behind the scenes?
OR
Notifying a user about intentionally ignoring an error is sufficient?

Also, I feel maybe we should not use `so we will create a new one`. Because who is we does not seem clear.

kubeflow/hack/create_context.sh
subodh101: Just omitting `Always` because it will make sense automatically every time when updating or installing.
Is that okay?

kubeflow/hack/create_context.sh
```diff
@@ -6,24 +6,18 @@
...
-
-RESULT=$?
-
-if [ ${RESULT} -eq 0 ]; then
-echo kubeconfig context ${NAME} already exists
-exit 0
-fi
+# Delete the existing kubeconfig, because it may be an outdated
+# context to a deleted cluster with the same name.
+kubectl config delete-context ${NAME} || echo "Context ${NAME} doesn't exist, Rename the context step will create it"
```
Bobgy, line 13: > Does a user really need to get the details of what is going to happen behind the scenes?
OR
Notifying a user about intentionally ignoring an error is sufficient?

We are going to change an external state (kubeconfig) that Kubeflow doesn't manage, sounds like we should better tell the user about it.

> Also, I feel maybe we should not use so we will create a new one. Because who is we does not seem clear.

That's a good point!

Bobgy (commented):
@jlewi what do you think?

My main support for this change is that, if we defer to documentation, many users may take a much longer time to find out the problem (because it's hard to find the exact documentation by error message).

This PR fixes it, and there're no foreseeable problems so far.

jlewi:
@Bobgy Your call.

subodh101:
> My main support for this change is that, if we defer to documentation, many users may take a much longer time to find out the problem (because it's hard to find the exact documentation by error message).

I thought about fixing this problem in kubectl.
However:
- that may take time.
- even if fixed, we will have to tell explicitly to the users the right kubectl client versions to use.

If we don't do it this way and let's say user is able to identify the problem then the user will do it manually(error-prone).

Bobgy:
/test all
Thanks! Then let's get this in.

Issue #115, opened 2020-08-06
Add VirtualService to allow GCLB health checks and get rid of backend updater statefulset


jlewi:
We currently use this stateful set as a hack to set the health check for the GCLB
https://github.com/kubeflow/manifests/tree/b3f5b41780a01845619884e638f4f6c9183d8fef/gcp/iap-ingress/base

It looks like there is now a better approach based on creating a virtual service which will just match the health check requests and reroute them to the readiness probe.
https://cloud.google.com/solutions/integrating-https-load-balancing-with-istio-and-cloud-run-for-anthos-deployed-on-gke

```
apiVersion: networking.istio.io/v1alpha3
kind: VirtualService
metadata:
  name: health  
spec:
  gateways:
  - istio-system/ingressgateway
  hosts:
  - "*"
  http:
  - match:
    - headers:
       user-agent:
         prefix: GoogleHC
      method:
       exact: GET
      uri:
        exact: /
    rewrite:
      authority: istio-ingressgateway.istio-system.svc.cluster.local:15020
      uri: /healthz/ready
    route:
    - destination:
        host: istio-ingressgateway.istio-system.svc.cluster.local
        port:
          number: 15020
```

I tried it and it works. 

Getting rid of the backend-updater pod should improve reliability; we've had various flakiness issues related to the healthcheck not getting set correctly.

jlewi:
To make this work we will also need to figure out how to exclude the health checks from the JWT checking.

It looks like JwtTriggerRule only allows matching paths.
https://istio.io/v1.1/docs/reference/config/istio.authentication.v1alpha1/#Jwt-TriggerRule

One option would be to exclude "/"  and then add JWT validation on the centraldashboard service.

Perhaps when we upgrade to ISTIO 1.6 we will have more options; e.g. we can match on headers.

jlewi:
Actually looks like BackendConfig will allow us to explicitly set the health check path.
https://cloud.google.com/kubernetes-engine/docs/how-to/ingress-features

Although it looks like that is beta in 1.17.6-gke.11 . Regular channel is still on 1.16

PR #141, opened 2020-10-11
Explicitly disable container native load balancing (neg) 

fabito:
# Problem

`envoy-ingress` backend services unhealthy when installing Kubeflow on GKE version `1.17.9-gke.1504`

# Solution

Add the `cloud.google.com/neg: '{"ingress": false}'` annotation to `istio-ingressgateway` to disable **container-native load balancing** (neg) which is enabled by default for GKE clusters >= 1.17 ( https://cloud.google.com/kubernetes-engine/docs/how-to/container-native-load-balancing#using )

### DIFF
--- a/kubeflow/instance/kustomize/iap-ingress/kustomization.yaml
+++ b/kubeflow/instance/kustomize/iap-ingress/kustomization.yaml
@@ -3,6 +3,7 @@ kind: Kustomization
 patchesStrategicMerge:
 - iap-ingress-config.yaml
 - service-accounts.yaml
+- service.yaml
 resources:
 - ../../../upstream/manifests/gcp/iap-ingress/v3 # {"$kpt-set":"gcp-iap-ingress-v3"}
 - ../../../upstream/manifests/istio/iap-gateway/base # {"$kpt-set":"istio-iap-gateway-base"}
--- a/kubeflow/instance/kustomize/iap-ingress/kustomization.yaml
+++ b/kubeflow/instance/kustomize/iap-ingress/kustomization.yaml
@@ -3,6 +3,8 @@ kind: Kustomization
 patchesStrategicMerge:
 - iap-ingress-config.yaml
 - service-accounts.yaml
+# Explicitly disable container native load balancing
+# See: https://github.com/kubeflow/gcp-blueprints/pull/141
 - service.yaml
 resources:
 - ../../../upstream/manifests/gcp/iap-ingress/v3 # {"$kpt-set":"gcp-iap-ingress-v3"}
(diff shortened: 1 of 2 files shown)


kubeflow/instance/kustomize/iap-ingress/kustomization.yaml
```diff
@@ -3,6 +3,7 @@ kind: Kustomization
 patchesStrategicMerge:
 - iap-ingress-config.yaml
 - service-accounts.yaml
+- service.yaml
 resources:
 - ../../../upstream/manifests/gcp/iap-ingress/v3 # {"$kpt-set":"gcp-iap-ingress-v3"}
```
Bobgy, line 8: Would you mind contributing to this upstream manifest instead of here?
Because instance folder should only include personal customizations

Bobgy (commented):
Thank you for the fix @fabito!
Can you refer to some reference why container native LB breaks sth?

kubeflow/instance/kustomize/iap-ingress/kustomization.yaml
fabito: @Bobgy I created this other PR https://github.com/kubeflow/manifests/pull/1579

I'm not familiar how the unit tests work. Let me know if I need to fix / adjust anything

fabito:
> Thank you for the fix @fabito!
> Can you refer to some reference why container native LB breaks sth?

Sorry @Bobgy, I'm not familiar with NEG, so I don't exactly what went wrong. The only evidence I had was my unhealthy back end (which is fixed now).

After disabling neg I just deleted the `iap-enabler`.

jlewi:
@fabito Thanks for reporting the issue and submitting a fix.

I wonder if this is an issue with health checks and the NEG is a red herring.
https://cloud.google.com/kubernetes-engine/docs/how-to/container-native-load-balancing#using

See #115

Starting with 1.17 BackendConfig should allow us to explicitly set the health check. I wonder if that would fix the issue?

@Bobgy  I think there is a couple paths here

1. We could check in the fix and then do further investigation later to undo it
1. We could file an issue to track supporting 1.17 and mention this as a problem with a possible work around.

I think its your call. I probably lean more towards the 2nd option.

Bobgy:
@jlewi Thank you!

1.17 has been rolled out to regular channel, so I d prefer option 1 here.

jlewi:
@Bobgy SGTM; @Bobgy Do you want to review and approve this then.

/ok-to-test

kubeflow/instance/kustomize/iap-ingress/kustomization.yaml
```diff
@@ -3,6 +3,7 @@ kind: Kustomization
 patchesStrategicMerge:
 - iap-ingress-config.yaml
 - service-accounts.yaml
+- service.yaml
```
Bobgy, line 6: nit: can we add a comment here with short explanation of why we want to add this patch and linking to the issue?

Bobgy (commented):
Because this is a temporary workaround with current knowledge, sounds good keeping it in instance folder.

PR #155, opened 2020-11-13
feat: install and upgrade instructions for management cluster in Kubeflow 1.2

Bobgy:
/assign @chensun

PR #156, opened 2020-11-14
Fix kubeflow makefile test

subodh101:
What?
Setting ZONE variable in the hydrate test is missing.

Since we are using the values from the instance to the upstream.
We don't need to set for upstream.
Failing test: https://github.com/kubeflow/gcp-blueprints/pull/155

/cc @Bobgy


subodh101:
@googlebot I fixed it.

Issue #158, opened 2020-11-15
[Discussion] UX for kpt cfg set both instance and upstream folders


Bobgy:
Discussion thread: https://github.com/kubeflow/gcp-blueprints/pull/155/files#r523194024

Options:
1. Ask users call `kpt cfg set` for both `instance` and `upstream` folder for each variable
2. `instance/settings.yaml` stores dictionary of settings, and set upstream folder using it
3. `instance/Kptfile` stores kpt setters and set upstream folder using it

In Kubeflow 1.1, we used option 1.
Right now, we have migrated to option 3 based on discussion in #120 .

@chensun raised some concerns this is not very straightforward to users.

> Personally, I feel it's much easier to open a file and edit three places in comparison to running three commands. 
While I realize this `./instance/Kptfile` looks a bit complex, the `./instance/settings.yaml` existed in the early version looked very straightforward and easy to be edited directly. Why did we drop that file?

_Originally posted by @chensun in https://github.com/kubeflow/gcp-blueprints/pull/155#r523194024_

Arguments preferring option 3:
> I think using kpt consistently is preferable as this creates a consistent experience. 
Users can do `kpt cfg list-setters .` to see all of the setters and then set them appropriately. 
This is consistent with respect to how arbitrary kpt packages are managed.
e.g. if a user downloads some new kpt package the expectation would be they run list-setters to see what setters exist and then set them.
>
> In contrast settings.yaml was a non-standard, KF specific solution. So I think getting rid of that in favor of standardizing on Kpt is preferable.

_Originally posted by @jlewi in https://github.com/kubeflow/gcp-blueprints/pull/155#r523434109_

Bobgy:
/cc @jlewi @chensun @subodh101 
Let's use this issue for further discussion, because this change is not included in scope of https://github.com/kubeflow/gcp-blueprints/pull/155

Bobgy:
For goals of the UX, I agree with the following:
1. We do not want to hide `kpt` semantics or build an abstraction on top
2. We want to avoid asking users setting the same values twice

When we evaluate the goals, I feel like option 3 isn't fully achieving that either, because
reading our instructions, users would think they use `kpt cfg set` on instance folder to set something
and then something magically set all the values on the upstream folder (under the assumption they never looked into the Makefile).

Ideally, I think we want them to know that, they are just using `kpt cfg set` to configure both instance and upstream folder. The situation is that both folders have the same parameter set.

Hmmmm, can we do this instead?
[...]

Issue #90, opened 2020-07-17
Better support for upgrading the deployment process - use Tekton and Kfctl instead of (in addition to) Make 


jlewi:
When we transitioned to blueprints for GCP we stopped using kfctl. There were a bunch of reasons that I won't go into here. A big one was that it was faster to iterate just by using a Makefile to script the various kustomize, kpt, and kubectl commands needed to describe the deployment process. The resulting Makefile's are also fairly readable and users can easily customize them.

An unresolved question is how we make it easy for users to pull in updates to the deployment recipe.

Right now a Kubeflow install is laid out as follows

```
my-kubeflow/
    Makefile
    instance/...
    upstream/...
```

To upgrade a user just re-places upstream with the latest copy and then redeploys. We'd like the same process to pull in any updates to the deployment process. For example, if we've added extra steps to hydrate or to apply we'd like that to be applied.

The problem right now is that the the deployment recipe is stored in the Makefile which is outside of upstream.

The deployment recipe is the sequence of kubectl, kpt, istioctl commands to be executed to produce the manifests and apply them.

We'd like to store the recipe in upstream so that the recipe gets updated when a user pulls in upstream.

While we could continue to use Makefile to define the recipe, a more cloud native way would be to use a YAML file listing out the commands. Tekton pipeline's and Tasks provide an existing resource that is well suited for this purpose. For example, we could easily chunk up our Makefile into tasks (e.g. hydrate and apply) that would contain steps corresponding to the commands currently invoked by the Makefile.

The downside of Tekton is that it requires Tekton and thus a K8s cluster to run. 

This could be addressed by a CLI that would simply read in the Tekton resources and execute the commands by shelling out. 

We could easily bake this functionality into kfctl; e.g.

```
kfctl apply -f deploy-kf-pipeline.yaml
```

@kkasravi suggested something along these lines awhile back.

jlewi:
@animeshsingh FYI an idea for how we might re-integrate kfctl into the GCP process.

Bobgy:
I worry that may be complicating things more than necessary. A local executor for tekton steps is also something new and we'd need documentation for which features we support and which not.

Did we consider just writing bash scripts for each make target and put those bash scripts in upstream?

Bobgy:
For needed information, we can set up a convention of using env vars to pass to these arguments or a config.yaml in a conventional location.
I think both works, I'd probably prefer a config.yaml.

Bobgy:
[...]

PR #163, opened 2020-11-16
feat: refactor management/Makefile to make upgrade easier. Part of #90 and #158

Bobgy:
Part of https://github.com/kubeflow/gcp-blueprints/issues/90 and https://github.com/kubeflow/gcp-blueprints/issues/158.

Changes:
1. make future upgrading management package as simple as `make update`, even there are makefile changes
2. better separation of concerns
    * add a `packages/management/Makefile` that can be used to deploy the management package directly
    * but it can also be configured as a recursive Makefile in an upstream package
3. update docs to reflect current status
4. add a `set-values.sh` script where users can set their values

All of the above changes are completely compatible with existing `instance` folder setup, so users won't need to touch `instance` folder during upgrade.


Bobgy:
/assign @chensun @jlewi @subodh101 
for review

management/Makefile
```diff
@@ -1,132 +1,76 @@
...
-# Directory where manifests should be fetched to
-MANIFESTS_DIR=./upstream/management
+PACKAGE_URL?=https://github.com/kubeflow/gcp-blueprints.git/packages/management@master
 
 INSTANCE_DIR=./instance
 
+# Directory where packages should be read from. Export these variables, so
+# recursive make commands can get them.
+export PACKAGE_DIR=./upstream/management
+export PACKAGE_CLUSTER_DIR=$(INSTANCE_DIR)/cluster
```
chensun, line 18: For my education, why do we have a variable for instance folder but not for `./upstream` for instance? Will there be multi-instance support?

management/set-values.sh
```diff
@@ -0,0 +1,22 @@
...
+#
+# The values you set will be stored in ./instance/Kptfile and
+# ./upstream/management/Kptfile, so they will be preserved if you commit the
+# changes to source control.
+
+set -ex
+
+for package in (./instance ./upstream/management)
+do
+    kpt cfg set $package name $NAME
```
chensun, line 19: nit: prefer quoting variables (ref: https://google.github.io/styleguide/shellguide.html#variable-expansion)

chensun (commented):
/lgtm

management/Makefile
Bobgy: Because the instance folder is just an empty kustomization.yaml that points to the upstream folder.

Bobgy:
Hmmm, the new auto approval is kind of confusing.

Feel free to add more commnents and I'll address them in a separate PR.

management/set-values.sh
Bobgy: Thanks, will do

PR #166, opened 2020-11-18
feat(management): update instructions to use kpt cfg set -R to set both packages. Part of #158

Bobgy:
/assign @subodh101 @jlewi @chensun 
for review

Part of https://github.com/kubeflow/gcp-blueprints/issues/158.

Changes: this PR consolidates discussion in https://github.com/kubeflow/gcp-blueprints/issues/158 and apply the result to management cluster instructions. Basically, we use `kpt cfg set -R` to set multiple packages at once.

If everything looks good, I'll be applying the same principles to kubeflow cluster in a separate PR.


subodh101 (approved):
/lgtm
It is great 🚀.
kpt version used: `0.36.1`.

Issue #139, opened 2020-09-23
Kubeflow on GCP - 1.2.0 release tracker


Bobgy:
Creating an issue to track efforts required before 1.2.0 release.

EDIT: upstream Kubeflow 1.2 release issue: https://github.com/kubeflow/kubeflow/issues/5371

TODOs:
* [x] P0 https://github.com/kubeflow/gcp-blueprints/issues/143
* [x] P0 Update KFP to 1.0.4
* [x] P0 https://github.com/kubeflow/gcp-blueprints/pull/169
* [x] Upgrade config connector to latest https://github.com/kubeflow/manifests/pull/1595
* [x] P0 https://github.com/kubeflow/gcp-blueprints/issues/145
* [x] P0 https://github.com/kubeflow/gcp-blueprints/issues/142 we need documentation
* [x] P1 Adopt config connector GKE add-on or operator installation: https://github.com/kubeflow/gcp-blueprints/issues/148
* [x] P1 https://github.com/kubeflow/gcp-blueprints/issues/149
* [x] P1 https://github.com/kubeflow/gcp-blueprints/issues/160
* [x] P1 kubeflow/gcp-blueprints#155
* [x] P1 usability: tell troubleshooting instructions when GCP resources timed out https://github.com/kubeflow/gcp-blueprints/pull/171, validate characters in name https://github.com/kubeflow/gcp-blueprints/pull/170
* [x] P1 https://github.com/kubeflow/gcp-blueprints/pull/173

Out of scope
* [ ] P2 https://github.com/kubeflow/gcp-blueprints/issues/158 (there's a blocker)
* [ ] P2 https://github.com/kubeflow/gcp-blueprints/issues/144
* [ ] P2 https://github.com/kubeflow/pipelines/issues/4356
* [ ] P2 https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/259#issuecomment-681145805 change nodeConfig to separate nodePool

subodh101:
[Website doc updates] Update setting the values step: https://github.com/kubeflow/website/pull/2126

Bobgy:
Renamed to 1.2.0 release tracker, because we are close to Kubeflow 1.2 release.

Bobgy:
https://github.com/kubeflow/gcp-blueprints/pull/169 is the last PR we need to get in. After that, I believe we can cut v1.2.0 release once the manifest repo pins a version.

Bobgy:
[...]

PR #168, opened 2020-11-20
chore(release): pin kubeflow/manifests v1.2-branch

Bobgy:

We should pin v1.2-branch to verify branched content


Bobgy:
Force merge to start testing

PR #170, opened 2020-11-23
feat: validate cluster name by regex. Fixes #154

### LINKED ISSUE #154: Kubeflow Makefile should validate the value of `$(NAME)`
I set [`$(NAME)`](https://github.com/kubeflow/gcp-blueprints/blob/36e6d0ad28bb85f83fdbee4c18229b18e56e98c2/kubeflow/Makefile#L37) with value `kf1.1`, which turned out to be an invalid name, but there was no complaint from `make set-values`.

Later running `make apply` failed with error:
```
kubectl --context=kf-cluster-201112 wait --for=condition=Ready --timeout=600s  containercluster kf1.1
error: timed out waiting for the condition on containerclusters/kf1.1
```

Trouble shooting with `kubectl describe containercluster --context $MANAGEMENT -n $PROJECT kf1.1` revealed:
```
Events:
  Type     Reason              Age                 From                         Message
  ----     ------              ----                ----                         -------
  Warning  DependencyNotReady  11m (x10 over 66m)  containercluster-controller  reference IAMServiceAccount chesu-dev/kf1.1-vm is not ready
```

Next, `kubectl describe iamserviceaccount --context $MANAGEMENT -n $PROJECT kf1.1-vm` revealed:
```
Events:
  Type     Reason        Age                 From                          Message
  ----     ------        ----                ----                          -------
  Normal   Updating      13m (x21 over 69m)  iamserviceaccount-controller  Update in progress
  Warning  UpdateFailed  13m (x21 over 69m)  iamserviceaccount-controller  Update call failed: error validating configuration: "account_id" ("kf1.1-vm") doesn't match regexp "^[a-z](?:[-a-z0-9]{4,28}[a-z0-9])$"
```

## Conclusion:

Given we expect three service accounts formed with this name:
- `NAME-admin`
- `NAME-user`
- `NAME-vm`

We need to validate its value to make sure that:
1. its length doesn't exceed 24.
1. it must start with [a-z] and contain only [-a-z0-9].

Bobgy:

Fixes https://github.com/kubeflow/gcp-blueprints/issues/154
/assign @chensun @subodh101

### DIFF
--- a/kubeflow/Makefile
+++ b/kubeflow/Makefile
@@ -27,18 +27,20 @@ BUILD_DIR=.build
 # The URL you want to fetch manifests from
 MANIFESTS_URL?=https://github.com/kubeflow/manifests.git@v1.2-branch
 
+cluster_name_regex=^[a-z][-a-z0-9]{2,16}[a-z0-9]$$
 # Validate cluster values are changed from default dummy values
 .PHONY: validate-values
 validate-values:
+	@if echo '$(NAME)' | egrep '$(cluster_name_regex)' >/dev/null; then \
+		echo 'The kubeflow cluster name "$(NAME)" is valid.'; \
+	else \
+		echo 'The kubeflow cluster name "$(NAME)" may contain only lowercase alphanumerics and "-", must start with a letter and end with an alphanumeric, and no longer than 18 characters.'; \
+	fi
 ifeq ($(shell test "$(MGMTCTXT)" =  MANAGEMENT-CTXT  -o  \
                    "$(NAME)"     =  KUBEFLOW-NAME  -o  \
                    "$(LOCATION)" =  LOCATION  -o  \
                    "$(PROJECT)"  =  PROJECT  &&  printf "true"), true)
 	$(error Either of MGMTCTXT, NAME, LOCATION, PROJECT values not set)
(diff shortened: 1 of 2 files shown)


subodh101:
Thank you!
I think we can validate but should not stop the flow for the changes that we are not 100% sure about. Also because it may change in future.

For e.g.,
[Frontend](https://github.com/kubeflow/kubeflow/blob/master/components/crud-web-apps/common/frontend/kubeflow-common-lib/projects/kubeflow/src/lib/form/validators.ts#L105) is still validating names with dns1035 which is not letting the notebooks pods getting created with [rfc 1123](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/).

Bobgy:
Hi @subodh101, I believe we chose Makefile, because users can hopefully easily edit them and maintain by themselves if that actually happens (compared to like a go binary).

Bobgy:
You raised a good point though, just that I don't think that's worth the complexity to either:
* add an option built into the makefile to ignore this error
* make the rule a complete warning (and users may simply didn't see it)

Bobgy:
I'll self lgtm to start testing for release. Feel free to raise more comments later.

PR #171, opened 2020-11-23
feat: wait for all GCP resources and more helpful error message

Bobgy:

Most of the issues raised here were related to errors deploying GCP resources,
so I improved the script to verify those resources are ready.

changes:
* verify all google cloud resources are ready
* show an error message for how to troubleshoot when sth times out


Bobgy:
I'll self lgtm to start testing for release. Feel free to raise more comments later.

Issue #176, opened 2020-11-30
KFServing fail with "Bad response from application: 503" when sending a prediction request through ingress and IAP


jal06:
When trying KFServing [GCP/IAP example](https://github.com/kubeflow/kfserving/tree/master/docs/samples/gcp-iap) , I receive an error `Bad response from application: 503`

After some investigations (more details in issue [1199](https://github.com/kubeflow/kfserving/issues/1199) ), we discovered that the prediction request cannot reach the KFServing endpoint because the Istio local gateway is missing

Bobgy:
Quoted from #177 

> Hi @jal06, I think the PR is not the idiomatic way to install cluster-local-gateway with istio/asm (anthos service mesh).
> 
> Because looking at https://github.com/knative/docs/blob/master/docs/install/installing-istio.md#installing-istio-without-sidecar-injection (strange that it's different from https://knative.dev/docs/install/installing-istio/#installing-istio-without-sidecar-injection),
> there's a declarative configuration in IstioOperator that allows installing cluster-local-gateway:
> 
> ```
>   components:
>     ingressGateways:
>       - name: istio-ingressgateway
>         enabled: true
>       - name: cluster-local-gateway
>         enabled: true
>         label:
>           istio: cluster-local-gateway
>           app: cluster-local-gateway
>         k8s:
>           service:
>             type: ClusterIP
>             ports:
>             - port: 15020
>               name: status-port
>             - port: 80
>               targetPort: 8080
>               name: http2
>             - port: 443
>               targetPort: 8443
>               name: https
> ```
> 
> This seems to be the correct approach when installing istio using operator. Similarly, we are installing asm using the operator per definition in https://github.com/kubeflow/manifests/blob/master/gcp/v2/asm/istio-operator.yaml.
> Therefore, we should include a similar ingressGateway definition in that operator. This will ensure the added cluster-local-gateway work with the entire service mesh.
> 
> and I suspect that installing cluster-local-gateway using kubeflow/manifests installed a gateway with old istio versions not designed for ASM, that's why it fails IAP authentication in your later experimentation.

Bobgy:
I think next steps would be
1. upgrade ASM to 1.6+ as per https://github.com/kubeflow/gcp-blueprints/issues/144
2. add the cluster-local-gateway definition during installation
3. verify whether kfserving works e2e

PR #177, opened 2020-11-30
fix(kfserving): add istio local gateway deployment in istio-system

jal06:
For solving issue [176](https://github.com/kubeflow/gcp-blueprints/issues/176)

### DIFF
--- a/kubeflow/Makefile
+++ b/kubeflow/Makefile
@@ -160,6 +160,9 @@ hydrate-kubeflow:
 	kustomize build --load_restrictor none -o $(BUILD_DIR)/metacontroller $(KF_DIR)/metacontroller
 	mkdir -p $(BUILD_DIR)/kubeflow-issuer
 	kustomize build --load_restrictor none -o $(BUILD_DIR)/kubeflow-issuer $(KF_DIR)/kubeflow-issuer
+	mkdir -p $(BUILD_DIR)/cluster-local-gateway
+	kustomize build --load_restrictor none -o $(BUILD_DIR)/cluster-local-gateway $(KF_DIR)/cluster-local-gateway
+
 
 
 # Hydrate resources to mirror images
@@ -219,6 +222,8 @@ apply-kubeflow: hydrate
 	kubectl --context=$(KFCTXT) apply -f ./$(BUILD_DIR)/application
 	kubectl --context=$(KFCTXT) apply -f ./$(BUILD_DIR)/cloud-endpoints
 	kubectl --context=$(KFCTXT) apply -f ./$(BUILD_DIR)/iap-ingress
+	kubectl --context=$(KFCTXT) apply -f ./$(BUILD_DIR)/cluster-local-gateway	
+
 
 	# Apply the namespace first
 	kubectl --context=${KFCTXT} apply -f ./$(BUILD_DIR)/knative/~g_v1_namespace_knative-serving.yaml
--- a/kubeflow/Makefile
+++ b/kubeflow/Makefile
@@ -161,7 +161,7 @@ hydrate-kubeflow:
 	mkdir -p $(BUILD_DIR)/kubeflow-issuer
 	kustomize build --load_restrictor none -o $(BUILD_DIR)/kubeflow-issuer $(KF_DIR)/kubeflow-issuer
 	mkdir -p $(BUILD_DIR)/cluster-local-gateway
-	kustomize build --load_restrictor none -o $(BUILD_DIR)/cluster-local-gateway $(KF_DIR)/cluster-local-gateway
+	kustomize build --load_restrictor none -o $(BUILD_DIR)/cluster-local-gateway $(MANIFESTS_DIR)/istio/cluster-local-gateway/base_v3
 
 
 
--- a/kubeflow/instance/kustomize/kustomization.yaml
+++ b/kubeflow/instance/kustomize/kustomization.yaml
@@ -1,4 +1,0 @@
-kind: Kustomization
-namespace: istio-system
-resources:
-- ../../../upstream/manifests/istio/cluster-local-gateway/base 
--- a/kubeflow/instance/kustomize/kustomization.yaml
+++ b/kubeflow/instance/kustomize/kustomization.yaml
@@ -1,0 +1,4 @@
+kind: Kustomization
+namespace: istio-system
+resources:
+- ../../../upstream/manifests/istio/cluster-local-gateway/base 
--- a/kubeflow/instance/kustomize/cluster-local-gateway/kustomization.yaml
+++ b/kubeflow/instance/kustomize/cluster-local-gateway/kustomization.yaml
@@ -1,0 +1,5 @@
+kind: Kustomization
+namespace: istio-system
+resources:
+- ../../../upstream/manifests/istio/cluster-local-gateway/base 
+
--- a/kubeflow/instance/kustomize/cluster-local-gateway/kustomization.yaml
+++ b/kubeflow/instance/kustomize/cluster-local-gateway/kustomization.yaml
@@ -2,4 +2,3 @@ kind: Kustomization
 namespace: istio-system
 resources:
 - ../../../upstream/manifests/istio/cluster-local-gateway/base 
-
--- a/kubeflow/instance/kustomize/cluster-local-gateway/kustomization.yaml
+++ b/kubeflow/instance/kustomize/cluster-local-gateway/kustomization.yaml
@@ -1,4 +1,0 @@
-kind: Kustomization
-namespace: istio-system
-resources:
-- ../../../upstream/manifests/istio/cluster-local-gateway/base 


kubeflow/instance/kustomize/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+kind: Kustomization
+namespace: istio-system
+resources:
+- ../../../upstream/manifests/istio/cluster-local-gateway/base
```
Bobgy, line 4: Hi @jal06, if the resource doesn't need any user modification, we can skip putting it in `instance` folder. You can just call `kustomize build` from the upstream folder directly.

Bobgy (commented):
Thank you for the fix!

jal06:
@Bobgy,

Please could you elaborate more ?
From my understanding, 
- the step `hydrate-kubeflow` does a copy of the manifests from `instance` to `upstream`.
- The step `apply-kubeflow` does a copy of the manifests from `upstream` to `.build` before applying the manifests from `.build`

If I just call kustomize build in the step `apply-kubeflow` without hydrating `cluster-local-gateway`, I get the error below during the step `make apply`

```
error: the path "./.build/cluster-local-gateway" does not exist
Makefile:191: recipe for target 'apply-kubeflow' failed
make: *** [apply-kubeflow] Error 1
```

That's why I don't understand how to skip putting `cluster-local-gateway` in instance.

kubeflow/Makefile
```diff
@@ -160,6 +160,9 @@ hydrate-kubeflow:
 	kustomize build --load_restrictor none -o $(BUILD_DIR)/metacontroller $(KF_DIR)/metacontroller
 	mkdir -p $(BUILD_DIR)/kubeflow-issuer
 	kustomize build --load_restrictor none -o $(BUILD_DIR)/kubeflow-issuer $(KF_DIR)/kubeflow-issuer
+	mkdir -p $(BUILD_DIR)/cluster-local-gateway
+	kustomize build --load_restrictor none -o $(BUILD_DIR)/cluster-local-gateway $(KF_DIR)/cluster-local-gateway
```
Bobgy, line 164: @jal06 I was suggesting changing this line to
```
kustomize build --load_restrictor none -o $(BUILD_DIR)/cluster-local-gateway $(MANIFESTS_DIR)/path/to/cluster-local-gateway
```
because there's no need to add user customizations to that package.

Bobgy:
Let me clarify what the two steps are doing.

> * the step `hydrate-kubeflow` does a copy of the manifests from `instance` to `upstream`.

The step `hydrate-kubeflow` runs kustomize on `instance/*` folders (right now) and generates hydrated manifests to `.build/*`. (but it doesn't matter where the source of the kustomize folder lives, we can change each kustomize command)

> * The step `apply-kubeflow` does a copy of the manifests from `upstream` to `.build` before applying the manifests from `.build`

The step `apply-kubeflow` applies resources in `.build/*` to your cluster.

kubeflow/Makefile
jal06: @Bobgy, 

I didn't test it, but I am not sure it could work because `make apply` starts by cleaning `.build` so cluster-local-gateway would be cleaned before being applied.

```
.PHONY: apply
apply: clean-build check-name check-iap apply-gcp wait-gcp create-ctxt apply-asm apply-kubeflow iap-secret
```
...
```
clean-build:
	# Delete build because we want to prune any resources which are no longer defined in the manifests
	rm -rf $(BUILD_DIR)/
	mkdir -p $(BUILD_DIR)/
```


Moreover, IMHO, when putting cluster-local-gateway in instance, the process of cluster-local-gateway is consistent with the process of other applications (kubeflow-issuer, metacontroller, kubeflow, ...) so it could be easier to maintain.
WDYT ?

jal06: @Bobgy, I am sorry, I missed th commit of cluster-local-gateway/kustomization.taml. I think it is OK now

Bobgy:
presubmit test is failing, I'll try to fix it next week
See https://github.com/kubeflow/gcp-blueprints/pull/180

kubeflow/Makefile
Bobgy: yes, consistency is a good point. Let me think about it more.

I was previously suggesting directly refer to upstream instead, because everything in `instance` folder needs user's manual maintenance (especially when upgrading), so if something doesn't need any customization, I preferred not putting it in instance folder.

jal06: > if something doesn't need any customization, I preferred not putting it in instance folder.

Agree; that makes sense
But IMHO deleting build because we want to prune any resources which are no longer defined in the manifests makes also  sense and I don't see any easy solution to do it. 

> Let me think about it more.

OK; thanks

Bobgy: @jal06
I guess you did not understand how this works
```
kustomize build --load_restrictor none -o $(BUILD_DIR)/cluster-local-gateway $(MANIFESTS_DIR)/path/to/cluster-local-gateway
```
Note that the `-o $(BUILD_DIR)/cluster-local-gateway` part is the output, and the `$(MANIFESTS_DIR)/path/to/cluster-local-gateway` part is the input.

I suggested changing `$(MANIFESTS_DIR)/path/to/cluster-local-gateway` to `....upstream/path/to/cluster-local-gateway`, the command will hydrate it everytime you apply to build folder. So that essentially means changing source of the hydration from instance to upstream, but build folder will always be hydrated each time. There won't be a problem there.

jal06:
@Bobgy,

Thanks for your explanations.
I think it is OK now (and I understand better how it works).

FYI, I did a test with Kubeflow 1.2.
In version 1.2, there is a new manifest in [base_v3 subfolder](https://github.com/kubeflow/manifests/tree/master/istio/cluster-local-gateway/base_v3)  for cluster-local-gateway (issue [1426](https://github.com/kubeflow/manifests/pull/1426) ).
So I also kustomized using kustomization.yaml base_v3.

Everything looks good but I get an error "`error while evaluating the ingress spec: BackendConfig istio-system/iap-backendconfig is not valid: error retrieving secret kubeflow-oauth: secrets "kubeflow-oauth" not found; BackendConfig istio-system/iap-backendconfig is not valid: error retrieving secret kubeflow-oauth: secrets "kubeflow-oauth" not found`"

Maybe this error is because I kustomized cluster-local-gateway from base_v3 ? 
Do we need to kustomize from base or from base_v3 ?
WDYT ?

CaptureErrEnvoyIngress

jal06:
Humm, after a while (several minutes) the error above disappeared. Everything is green

Bobgy:
Thanks! base_v3 is correct

Bobgy:
I'll fix the test infra failure before merging this PR

Bobgy:
Hi @jal06, I think the PR is not the idiomatic way to install cluster-local-gateway with istio/asm (anthos service mesh).

Because looking at https://github.com/knative/docs/blob/master/docs/install/installing-istio.md#installing-istio-without-sidecar-injection (strange that it's different from https://knative.dev/docs/install/installing-istio/#installing-istio-without-sidecar-injection), 
there's a declarative configuration in IstioOperator that allows installing cluster-local-gateway:
```
  components:
    ingressGateways:
      - name: istio-ingressgateway
        enabled: true
      - name: cluster-local-gateway
        enabled: true
        label:
          istio: cluster-local-gateway
          app: cluster-local-gateway
        k8s:
          service:
            type: ClusterIP
            ports:
            - port: 15020
              name: status-port
            - port: 80
              targetPort: 8080
              name: http2
            - port: 443
              targetPort: 8443
              name: https
```

This seems to be the correct approach when installing istio using operator. Similarly, we are installing asm using the operator per definition in https://github.com/kubeflow/manifests/blob/master/gcp/v2/asm/istio-operator.yaml.
Therefore, we should include a similar ingressGateway definition in that operator. This will ensure the added cluster-local-gateway work with the entire service mesh.

and I suspect that installing cluster-local-gateway using kubeflow/manifests installed a gateway with old istio versions not designed for ASM, that's why it fails IAP authentication in your later experimentation.

Bobgy:
but a problem I realized is that, asm installation is at 1.4 version currently, and in that version istio was using a different spec for istio operator definition file. I think we should probably first figure out asm upgrade path and directly use the latest format.

jal06:
@Bobgy, thanks for your feedback.
I am happy to know why my later experimentation fails with IAP authentication.

Do you think it will be possible to upgrade asm and set the idiomatic way to install cluster-local-gateway for the next version of Kubeflow (1.3 ) ?
I am not sure  I could help for that (lack of expertise and time) but I would be more than happy to test it and use it when this will be available

About asm version, I confirm I am using 1.4.10 (the recommended version in the documentation).

BTW, when trying to debug kfserving, I realised that Kiali could help a lot; but Kiali is not installed by default. Is it possible to add Kiali by default in the manifests ?

```
istioctl version
client version: 1.4.10-asm.18
cluster-local-gateway version: 
cluster-local-gateway version: 
galley version: 1.4.10-asm.15
galley version: 1.4.10-asm.15
ingressgateway version: 1.4.10-asm.15
ingressgateway version: 1.4.10-asm.15
nodeagent version: 
nodeagent version: 
pilot version: 1.4.10-asm.15
pilot version: 1.4.10-asm.15
sidecar-injector version: 1.4.10-asm.15
sidecar-injector version: 1.4.10-asm.15
data plane version: 1.4.10-asm.15 (20 proxies), 1.1.6 (2 proxies)
```

Bobgy:
> Do you think it will be possible to upgrade asm and set the idiomatic way to install cluster-local-gateway for the next version of Kubeflow (1.3 ) ?

Yes, at least upgrading asm will be on the roadmap, and it should be a lot easier to add `cluster-local-gateway` by then

Bobgy:
For Kiali, it sounds like an interesting tool to use. I haven't tried myself. In general, I think we'll only provide the minimal to let people run Kubeflow, they can choose to add tools they like

jal06:
@Bobgy, about Kiali, what you propose sounds relevant with the new policy of addons integration in Istio project : https://istio.io/latest/blog/2020/addon-rework/

Thanks !!!

Bobgy:
Let's close this for now and I'll follow up during next release.

edi-bice-by:
@Bobgy: "Let's close this for now and I'll follow up during next release."

I deployed KF 1.3 using gcp-blueprints on GCP with IAP and am having a hard time trying to get anything kfserving (both from inside and outside the cluster) working. 

Is the cluster-local-gateway still required?

zijianjoy:
> Is the cluster-local-gateway still required?

Hello @edi-bice-by , there are quite some changes on KFServing and its dependencies. KNative Serving has been upgraded to 0.22.0 to support KFServing, see requirement: https://github.com/kubeflow/kfserving#prerequisites. As the link described, `Since Knative v0.19.0 cluster local gateway deployment has been removed and shared with ingress gateway, if you are on Knative version later than v0.19.0 and KFServing version older than v0.6.0 you should modify localGateway to knative-local-gateway and localGatewayService to knative-local-gateway.istio-system.svc.cluster.local in the inference service config.`

edi-bice-by:
@zijianjoy yes, thank you, I observed the difference in naming and modified accordingly, though was surprised that the blueprint which sources precisely just that combination described in the documentation is not doing the correct thing in the first place. 

I'm still unable to query any of the kfserving gcp_iap samples (flower or sklearn) externally. More concerning, hence the post here, is that I cannot seem to be able to resolve hostnames (and query kfserving endpoints) from within the cluster either.

zijianjoy:
@edi-bice-by 

Although it is out-of-date, you might be able to find some useful information about querying KFServing endpoint using this sample: https://github.com/kubeflow/kfserving/tree/master/docs/samples/gcp-iap. For my attempt to make this endpoint available externally, see https://github.com/kubeflow/gcp-blueprints/issues/209#issuecomment-818027778. 

More specifically, I made relatively big change in this file for authorization https://github.com/kubeflow/kubeflow/blob/master/docs/gke/iap_request.py. The way I made the change is by creating desktop app client ID and client secret: https://cloud.google.com/iap/docs/authentication-howto#authenticating_from_a_desktop_app.

PR #179, opened 2020-12-08
feat: example instance folder to patch GCP resources

Bobgy:
/assign @chensun


kubeflow/instance/gcp_config/cluster-patch.yaml
```diff
@@ -0,0 +1,29 @@
...
+apiVersion: container.cnrm.cloud.google.com/v1beta1
+kind: ContainerCluster
+metadata:
+  name: KUBEFLOW-NAME # {"$kpt-set":"name"}
+# To replace default nodepool with your own, uncomment the following two lines
+# and add "nodepool-example.yaml" to "resources" field of kustomization.yaml.
+# annotations:
+#   cnrm.cloud.google.com/remove-default-node-pool: "false"
+spec:
+  location: "LOCATION" # {"$kpt-set":"location"}
```
chensun, line 26: nit: quoting or not, let's keep it consistent for `KUBEFLOW-NAME` and `LOCATION`

(IIUC, this the the source, and value flows into `Kptfile`, right?)

kubeflow/instance/gcp_config/nodepool-example.yaml
```diff
@@ -0,0 +1,44 @@
...
+# you need to edit `./kustomization.yaml` and add `nodepool-example.yaml` in
+# resources field to include this.
+apiVersion: container.cnrm.cloud.google.com/v1beta1
+kind: ContainerNodePool
+metadata:
+  name: KUBEFLOW-NAME-example # {"$kpt-set":"node-pool-example"}
+# Refer to https://cloud.google.com/config-connector/docs/reference/resource-docs/container/containernodepool
+# for detailed fields and documentation.
+spec:
+  location: "LOCATION" # {"$kpt-set":"location"}
```
chensun, line 25: nit: consistent on quoting.

kubeflow/instance/gcp_config/nodepool-example.yaml
```diff
@@ -0,0 +1,44 @@
...
+  # nodeLocations:
+  # - gcp-zone-1
+  # - gcp-zone-2
+  initialNodeCount: 2
+  autoscaling:
+    minNodeCount: 2
+    maxNodeCount: 8
+  nodeConfig:
+    machineType: n1-standard-8
+    minCpuPlatform: 'Intel Broadwell'
```
chensun, line 36: nit: guessing no difference between `'` and `"`, pick a consistent one?

kubeflow/instance/gcp_config/kustomization.yaml
```diff
@@ -3,11 +3,20 @@
...
+namespace: PROJECT # {"$kpt-set":"project"}
 commonLabels:
   kf-name: KUBEFLOW-NAME # {"$kpt-set":"name-label"}
 resources:
+# The default Google Cloud resources.
 - ../../upstream/manifests/gcp/v2/cnrm
+# Add additional Google Cloud resources by including resources here.
+# For example, uncomment the following line to include an additional example
+# nodepool described in a local file ./nodepool-example.yaml.
+# - nodepool-example.yaml
```
chensun, line 17: (For my education) This feels like some sort of patching as well. What if the resource specified in this yaml already exists? Will it be replaced? merged? or there will be some error thrown?

Bobgy: If the resource already exists, then this will fail. User needs to switch to patchesStrategicMerge. Therefore, the intent is clear, if we include a resource, it's something new. If we add a patch, the base resource must exist.

kubeflow/instance/gcp_config/nodepool-example.yaml
Bobgy: I removed the quote, because it's also valid YAML

Bobgy:
Thank you @chensun !
I've addressed your comments

kubeflow/instance/gcp_config/kustomization.yaml
chensun: Thanks for the explanation!

chensun:
/lgtm

Thanks, Yuan!

Bobgy:
I'll try to fix the presubmit failure this week

PR #186, opened 2021-01-14
Update Kptfile

Bobgy:


Bobgy:
Per discussion with kpt team member, `isSet` should not be in the `Kptfile` in published packages

PR #193, opened 2021-01-27
Fix kubeflow-istio folder creation

kelvins:
Makefile is trying to create the `kubeflow-apps` folder twice.
The second time it should be `kubeflow-istio`.


Bobgy:
Ohhh, thank you so much!

Issue #209, opened 2021-04-10
[KF 1.3] Integrate Knative and KFServing


zijianjoy:
Highlevel tracking: https://github.com/kubeflow/manifests/issues/1798

In order to prepare for GCP distribution of Kubeflow 1.3, we need to make sure kfserving installation is working. Currently we haven't installed cluster-local-gateway in our deployment step, but it is needed based on documentation in [KFServing GCP/IAP Example](https://github.com/kubeflow/kfserving/tree/master/docs/samples/gcp-iap#expose-the-inference-service-externally-using-an-additional-istio-virtual-service). Based on manifest [README](https://github.com/kubeflow/manifests#knative), I ran the following command to build local-cluster-gateway:

```
kustomize build --load-restrictor LoadRestrictionsNone -o $(BUILD_DIR)/cluster-local-gateway ./common/istio/istio-1-9-0/cluster-local-gateway/base
```

The build result is in https://github.com/kubeflow/gcp-blueprints/commit/f2f05bb5c07012acfaaa31575a3235761bc24e82.

However, the cluster-local-gateway deployment is failing with the following error:

```
 warn Envoy proxy is NOT ready: config not received from Pilot (is Pilot running?): cds updates: 0 successful, 0 rejected; lds updates: 0 successful, 0 rejected
warn sds failed to warm certificate: failed to generate workload certificate: create certificate: rpc error: code = Unavailable desc = connection error: desc = "transport: authentication handshake failed: x509: certificate signed by unknown authority"
 warn ca ca request failed, starting attempt 1 in 96.070398ms
Info
 warn ca ca request failed, starting attempt 2 in 187.488321ms
Info
 warn ca ca request failed, starting attempt 3 in 429.265926ms
Info
 warn ca ca request failed, starting attempt 4 in 878.238921ms
```

How should I correctly configure certificate and CA? Since I am using anthos service mesh 1.9.2-1, the pilot is not a pod but lives inside istiod. See pods as below:

```
$ kubectl get pod -n istio-system
NAME                                     READY   STATUS    RESTARTS   AGE
backend-updater-0                        1/1     Running   0          39h
cluster-local-gateway-6b6cb58745-qq7r2   0/1     Running   0          7h7m
iap-enabler-5864d6c776-5s5lc             1/1     Running   0          39h
istio-ingressgateway-66c8544779-78z7n    1/1     Running   0          43h
istio-ingressgateway-66c8544779-vch9t    1/1     Running   0          43h
[...]

PR #210, opened 2021-04-15
(Batch Update) Prepare gcp-blueprint for Kubeflow 1.3 release

zijianjoy:
# Prepare gcp-blueprint for Kubeflow 1.3 release
 
## List of larger changes
 
1. Upgrade ASM version to 1.9.2-1 and corresponding Istio syntax
2. Refactor the whole repo structure to match upstream `kubeflow/manifests`
3. Alter the approach to fetch and use `kubeflow/manifests`. 
4. Decoupled kubeflow-apps to individual kustomization.yaml
5. Upgrade and install multiple kubeflow apps: Pipelines, KFServing, Volume, Tensorboard, Notebook, Katib, etc.
6. Use a custom version of profile controller and patched manifest
7. Apply Kustomize v4+ in the Makefile rule
 
## High level deployment steps
 
1. Assuming you have downloaded the necessary tools (Kustomize 4+, jq, yq, python, gcloud, gcloud components: kubectl kpt beta)
1. Pull from https://github.com/zijianjoy/gcp-blueprints/commits/match-upstream 
   * git clone https://github.com/zijianjoy/gcp-blueprints.git
   * cd gcp-blueprints
   * git fetch
   * git checkout match-upstream-share
1. $ cd kubeflow
1. $ bash ./pull-upstream.sh
1. Fulfill content for your kubeflow cluster in env.sh
1. $ source ./env.sh
1. $ bash ./kpt-set.sh
1. $ set CLIENT_ID and CLIENT_SECRET using IAP client from GCP console.
1. Set management cluster namespace
   * $ kubectl config use-context ${MGMTCTXT}
   * $ kubectl create namespace ${KF_PROJECT}
   * $ kubectl config set-context --current --namespace ${KF_PROJECT}
1. $ make apply-kubeflow
 
## Known issues
 
1. `make apply-kubeflow` might fail, rerunning the command multiple times will eventually resolve the issue. This should be able to resolve by using `kpt live apply`. But `kpt live apply` doesn't seem to understand some CRD like ContainerCluster.
1. KNative and KFserving need special installation: https://github.com/kubeflow/gcp-blueprints/issues/209. Those changes haven't been brought in yet. But will do later.
1. profile controller has several issues and pending PR: https://github.com/kubeflow/kubeflow/issues/5816, https://github.com/kubeflow/kubeflow/issues/5813, https://github.com/kubeflow/kubeflow/pull/5761. We need to configure profile controller correctly to include those fixes and patches.
1. ASM namespace label is currently hardcoded in multiple locations `istio.io/rev: asm-192-1`, but it is very easy that ASM version will be updated in the future. We need to figure out a way to configure them accordingly.
1. Many files under root directory haven't been cleaned up yet.
 
## To be determined 
 
1. We need to plan for kubeflow cluster upgrade path for future Kubeflow verison: upgrading individual component, adding component, upgrading common resources.
1. We need to determine the support tier and scope for Kubeflow on GCP: Do we support on KFP? Or do we support all kubeflow apps to be working on GCP? Or etc.
1. When kubeflow-apps/ is decoupled, we use `disableNameSuffixHash` to define kubeflow-config, many apps share this same config, we need to find a way to correctly reference the field in kubeflow-config. (clusterDomain and namespace)


zijianjoy:
cc @yanniszark for https://github.com/kubeflow/manifests/issues/1798

Bobgy:
Thank you!
/approve
We can only merge and iterate.

The OWNERS warning should better be resolved in upstream.

Bobgy:
/hold
fix owners

DavidSpek:
Regarding the plan for an upgrade path, I think my approach with the Argo CD deployment makes it easy for users and the maintainers to keep the components in their repository up to date with Renovate. Example for updating manifests as well as image tags: https://github.com/argoflow/argoflow/pull/17. Kubeflow manifest example: https://github.com/argoflow/argoflow/pull/2/files.

PR #215, opened 2021-04-18
(cluster) Use STABLE release channel, change machineType to e2-standard-8. Fix #213, Fix #214

### LINKED ISSUE #213: Pin cluster version to 1.18

### LINKED ISSUE #214: default to e2 machine type, instead of n1

zijianjoy:
Fix #213 
Fix #214

Prevent future deprecation of GCP dependencies, pin k8s version in Kubeflow 1.3 release.
N1 is going to be deprecated, we will use the equivalent machine type in E2.

### DIFF
--- a/kubeflow/common/cluster/upstream/cluster.yaml
+++ b/kubeflow/common/cluster/upstream/cluster.yaml
@@ -45,14 +45,23 @@ spec:
   releaseChannel:
     # Per https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/194
     # use upper case for the channels
-    channel: REGULAR
-  location: LOCATION # {"$kpt-set":"location"}
+    channel: UNSPECIFIED
+  nodeVersion: '1.18'
+  # Master version controls the kubernetes API version.
+  # Future upgrade of master version might break Kubeflow deployment.
+  # Therefore we need to pin master version and disable minor version upgrade.
+  # Autopilot will enforce automatic node upgrade.
+  minMasterVersion: '1.18' 
+  # This feature is very new: https://github.com/GoogleCloudPlatform/k8s-config-connector/releases/tag/v1.46.0
+  # If you encounter an issue that this field is unknown, remove this field from here for now.
+  enableAutopilot: false
+  location: location # {"$kpt-set":"location"}
   workloadIdentityConfig:
-    identityNamespace: PROJECT.svc.id.goog # {"$kpt-set":"identity-ns"}
+    identityNamespace: project-id.svc.id.goog # {"$kpt-set":"identity-ns"}
   loggingService: logging.googleapis.com/kubernetes
   monitoringService: monitoring.googleapis.com/kubernetes
   nodeConfig:
-    machineType: n1-standard-8
+    machineType: e2-standard-8
     metadata:
       disable-legacy-endpoints: "true"
     oauthScopes:
@@ -60,8 +69,6 @@ spec:
     - https://www.googleapis.com/auth/monitoring
     - https://www.googleapis.com/auth/devstorage.read_only
     serviceAccountRef:
-      name: KUBEFLOW-NAME-vm # {"$kpt-set":"name-vm"}
+      name: name-vm # {"$kpt-set":"name-vm"}
     workloadMetadataConfig:
       nodeMetadata: GKE_METADATA_SERVER
-  nodeVersion: '1.18'
-  minMasterVersion: '1.18'
--- a/kubeflow/common/cluster/upstream/cluster.yaml
+++ b/kubeflow/common/cluster/upstream/cluster.yaml
@@ -45,16 +45,15 @@ spec:
   releaseChannel:
     # Per https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/194
     # use upper case for the channels
-    channel: UNSPECIFIED
-  nodeVersion: '1.18'
+    channel: STABLE
   # Master version controls the kubernetes API version.
   # Future upgrade of master version might break Kubeflow deployment.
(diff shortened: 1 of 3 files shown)


packages/gcp-resources/cnrm/cluster/cluster.yaml
```diff
@@ -63,3 +63,6 @@ spec:
       name: name-vm # {"$kpt-set":"name-vm"}
     workloadMetadataConfig:
       nodeMetadata: GKE_METADATA_SERVER
+  nodeVersion: '1.18'
+  minMasterVersion: '1.18'
```
Bobgy, line 67: This looks weird, it says min master version, so can it be higher?

Bobgy: > The minimum version of the master. GKE will auto-update the master to new versions, so this does not guarantee the current master version--use the read-only master_version field to obtain that. If unset, the cluster's version will be set by GKE to the version of the most recent official release (which is not necessarily the latest version).

There is no guarantee, unfortunately.

https://cloud.google.com/config-connector/docs/reference/resource-docs/container/containercluster

zijianjoy: What we need to specify is the `nodeVersion`, however, it will produce validation error if we set `nodeVersion` but not set `minMasterVersion`. Therefore I set the `minMasterVersion` here. 

Master node is invisible to GKE user, and its version will be upgraded automatically, but we can see its version at the `status` -> `masterVersion` field. 

My understanding is that we need to pin the `nodeVersion` and disable node upgrade by setting `releaseChannel` to be `unspecified`. This combination makes sure that our dependencies on various GCP products won't be outdated, regardless of masterVersion. It is just that master version cannot be lower than nodeVersion, so ContainerCluster config requires me to specify minMasterVersion.

Bobgy: Makes sense, for 1.18 the major reason to stay is because of docker, so we want to pin node version.

However, for most other cases we care more about master version, because master version decides the Kubernetes API version. And releaseChannel is a config for Kubernetes master version.

So we need to check whether we need to disable auto node upgrade. Or can we enable node upgrade, but only allow upgrade until 1.18?

zijianjoy: Thank you Yuan for the insight! That is very helpful. 

Since master version is what we care about, I would suggest to pin the master using 1.18, because any major and minor version bump will require E2E validation from our side, and possibly break user experience. I am leaning towards such validation to happen only when there is a Kubeflow release. 

In terms of security, even if we choose `no channel`, security patch will still be applied https://cloud.google.com/kubernetes-engine/docs/concepts/release-channels#no_channel, because it is an infrastructure requirement. 

For auto node upgrade, if you agree that we should pin the version and subscribe to `no channel`, we should disable this field `enableAutopilot` at https://cloud.google.com/config-connector/docs/reference/resource-docs/container/containercluster. If we need to enable node upgrade, I tried to find such configuration which allows me to set maximum node version, but with no luck yet.

zijianjoy: Actually this feature is quite new : https://github.com/GoogleCloudPlatform/k8s-config-connector/releases/tag/v1.46.0

So when I apply this `enableAutopilot` flag, there is error: `Error from server: error when creating ".build/gcp_config/container.cnrm.cloud.google.com_v1beta1_containercluster_kf-ref4.yaml": admission webhook "deny-unknown-fields.cnrm.cloud.google.com" denied the request: unknown field "spec.enableAutopilot"`

Bobgy: Yes, it is. Can you confirm autopilot mode stage? Is it alpha or beta? We need to upgrade management cluster config connector to have access

zijianjoy: Although autopilot looks very new to me, it doesn't have an alpha or beta banner on documentation. Based on what David Spek mentioned, it doesn't support 3rd party webhook: https://github.com/jetstack/cert-manager/issues/3717.

We discussed offline and agreed to use STABLE release channel, we won't specify node/master version in cluster.yaml file (it will install 1.17.17-gke.3000 currently by default). I will update PR about this change. I am going to create a separate PR for config connector update as well.

kubeflow/common/cluster/upstream/cluster.yaml
```diff
@@ -45,14 +45,22 @@ spec:
   releaseChannel:
     # Per https://github.com/GoogleCloudPlatform/k8s-config-connector/issues/194
     # use upper case for the channels
-    channel: REGULAR
+    channel: STABLE
+  # Master version controls the kubernetes API version.
+  # Future upgrade of master version might break Kubeflow deployment.
+  # At the time of writing, STABLE release channel is using 0.17.
```
Bobgy, line 51: Nit: 1.17

Bobgy (commented):
/LGTM
/Hold
To fix a minor mistake

kubeflow/common/cluster/upstream/cluster.yaml
zijianjoy: Thank you Yuan for the good catch! Fixed.

zijianjoy:
/lgtm

Can I LGTM myself for the minor fix?

Bobgy:
/lgtm
FYI, you can add the LGTM label by yourself if you are confident the PR is ready to merge

zijianjoy:
> /lgtm
> FYI, you can add the LGTM label by yourself if you are confident the PR is ready to merge

Thank you for the information!

Issue #227, opened 2021-04-22
Add workload-identity-bindings-patch.yaml


zijianjoy:
Use KFP built-in TFX sample to verify permissions, it should be able to write to GCS

We need this for all apps, otherwise it might not work: https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#authenticating_to. profile controller not workiing

zijianjoy:
Validated that KFP and profiler are working correctly with workload identity

PR #229, opened 2021-04-22
(profiles, pipeline) Add workload identity binding for service account. Partial #227

zijianjoy:
Partial #227

Need some help validating GCS access after applying this change.

Profile controller is validated to be working to access gcp resources.


Bobgy:
/LGTM
Note, when we move workload identity bindings to each package, it's clearer that we might want to also move the corresponding GSA to its package. These actions continue to push each package towards installable by itself and not depend on common settings

zijianjoy:
> /LGTM
> Note, when we move workload identity bindings to each package, it's clearer that we might want to also move the corresponding GSA to its package. These actions continue to push each package towards installable by itself and not depend on common settings

Thank you Yuan! I added a card in `release 1.3` to track this change.

Issue #230, opened 2021-04-22
Simplify Makefile


zijianjoy:
### Motivation

We want to answer the following questions for users:

* How does user set kpt values?
* How do we define root Makefile for simplest logic.
* How does subfolder Makefile get common env variables from root Makefile?
* How does user pull from upstream while not interfering user kustomized items?
* How does user maintain the build folders?
* How does user upgrade the whole gcp-blueprints's Kubeflow manifest version? (1.3->1.4)
* How does user upgrade individual component? (For example they only want to upgrade KFserving)

### Thinking points

Some high level directions we are considering.

1. BUILD_DIR?=build for build folder path, default to each app own directory.
2. Polish deployment step.
3. individual app's apply instead of root Makefile make apply
4. call Make if Makefile exists, otherwise use kubectl.

zijianjoy:
Structure description:

### List of components:

common/cnrm  [special-handling]
common/asm  [special-handling]
iap-secret (Command only)
common/kubeflow-namespace
common/istio
contrib/application
common/cert-manager [special-handling]
contrib/metacontroller
common/cloud-endpoints
common/iap-ingress
common/config-kubeflow
apps/admission-webhook
apps/centraldashboard
common/kubeflow-roles
apps/jupyter
apps/volumes-web-app
apps/tensorboard/tensorboards-web-app [special-handling]
apps/tensorboard/tensorboard-controller [special-handling]
apps/profiles
apps/pytorch-job
apps/tf-training
apps/pipeline
apps/kfserving
common/user-namespace
apps/katib
common/cert-manager/upstream/cert-manager/kubeflow-issuer [special-handling]

### Root kubeflow structure

`env.sh`: User defines global environment variables
`kpt-set.sh`: User set custom values for kpt using environment variables
`pull-upstream.sh`: User set the upstream manifests version, and pull all manifests to each components' `upstream/` folder.
`Makefile`: Commands: validate-values, hydrate-all, apply-all, clean-build, delete-gcp.

#### hydrate-all

input: BUILD_DIR(default is ./build), list of component paths (default value is listed above in this comment)
logic: call validate-values, then call each components' `make hydrate` target, if Makefile not exists, call kustomize build. It generates resources file to the provided BUILD_DIR path (either relative or absolute path).

#### apply-all

input: BUILD_DIR(default is ./build), list of component paths (default value is listed above in this comment)
[...]

PR #234, opened 2021-04-26
(Makefile) Run make apply-v3 to apply all components and deployment steps. #230

zijianjoy:
Partial #230

This PR implements the core functionality of #230 - `make apply`

The way to initialize and deploy:

1. set all values in env.sh and kpt-set.sh
2. Set CLIENT_ID and CLIENT_SECRET
3. $ source env.sh
4. $ bash ./pull-upstream.sh
5. $ bash kpt-set.sh
6. $ make apply-v3

Rules like `hydrate`, `delete`, `clean-build` are not implemented yet. We need to remove existing rule in Makefile later on.

### DIFF
--- a/kubeflow/Makefile
+++ b/kubeflow/Makefile
@@ -1,3 +1,4 @@
+
 # The name of the context for the management cluster
 # These are read using yq from the Kptfile.
 MGMTCTXT=$(shell yq r ./Kptfile 'openAPI.definitions."io.k8s.cli.setters.mgmt-ctxt".x-k8s-cli.setter.value')
@@ -50,6 +51,67 @@ apply-cnrm:
 	rm -rf $(BUILD_DIR)/gcp_config && mkdir -p $(BUILD_DIR)/gcp_config
 	kustomize build --load-restrictor LoadRestrictionsNone -o $(BUILD_DIR)/gcp_config ./common/cnrm
 	kubectl --context=$(MGMTCTXT) apply -f ./$(BUILD_DIR)/gcp_config
+
+.PHONY: apply-component
+apply-component: validate-values check-name 
+ifeq ($(component_path),)
+	$(error component_path is not set for apply-component rule); 
+endif
+	echo 'Component path: $(component_path)' 
+	$(eval makefile_path = ./$(component_path)/Makefile)
+	echo 'Apply component resources: $(component_path)'
+	@if [ -f $(makefile_path) ] ; then \
+		echo 'Found Makefile, call `make apply` of this component Makefile.'; \
+		NAME=$(NAME) KFCTXT=$(KFCTXT) LOCATION=$(LOCATION) PROJECT=$(PROJECT) $(MAKE) -C $(component_path) apply build_dir=$(BUILD_DIR); \
+	else \
+		echo 'Makefile not found, use kustomize and kubectl to apply resources.'; \
+		component_build_dir=./$(component_path)/$(BUILD_DIR); \
+		rm -rf $$component_build_dir && mkdir -p $$component_build_dir; \
+		kustomize build --load-restrictor LoadRestrictionsNone -o $$component_build_dir ./$(component_path); \
+		kubectl --context=$(KFCTXT) apply -f $$component_build_dir; \
+	fi
+
+
+components= \
+common/cnrm \
+common/asm \
+common/iap-secret \
+common/kubeflow-namespace \
+common/istio \
+common/cert-manager \
+contrib/metacontroller \
+common/cloud-endpoints \
+common/iap-ingress \
+common/config-kubeflow \
+apps/admission-webhook \
+apps/centraldashboard \
+common/kubeflow-roles \
+apps/jupyter \
+apps/volumes-web-app \
+apps/tensorboard \
+apps/profiles \
+apps/pytorch-job \
+apps/tf-training \
+apps/pipeline \
+common/knative \
+apps/kfserving \
+common/user-namespace \
+apps/katib \
+common/kubeflow-issuer \
+
+
+.PHONY: apply-v3
+apply-v3: validate-values check-name 
+	@for component in $(components) ; do \
+	$(MAKE) apply-component component_path=$$component || break ; \
+	done
+
+	# Kick the IAP pod because we will reset the policy and need to patch it.
+	# TODO(https://github.com/kubeflow/gcp-blueprints/issues/14)
+	kubectl --context=$(KFCTXT) -n istio-system delete pods -l service=iap-enabler
+	# Kick the backend updater pod, because information might be outdated after the apply.
+	# https://github.com/kubeflow/gcp-blueprints/issues/160
+	kubectl --context=$(KFCTXT) -n istio-system delete pods -l service=backend-updater
 
 .PHONY: apply-kubeflow
 apply-kubeflow: validate-values check-name 
--- a/kubeflow/common/cert-manager/Makefile
+++ b/kubeflow/common/cert-manager/Makefile
@@ -1,0 +1,12 @@
+
+.PHONY: apply
+apply:
+	# Common cert-manager
+	rm -rf $(build_dir) && mkdir -p $(build_dir)
+	kustomize build --load-restrictor LoadRestrictionsNone -o $(build_dir) ./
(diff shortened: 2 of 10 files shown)


kubeflow/Makefile
```diff
@@ -51,6 +52,67 @@ apply-cnrm:
...
+	@if [ -f $(makefile_path) ] ; then \
+		echo 'Found Makefile, call `make apply` of this component Makefile.'; \
+		NAME=$(NAME) KFCTXT=$(KFCTXT) LOCATION=$(LOCATION) PROJECT=$(PROJECT) $(MAKE) -C $(component_path) apply build_dir=$(BUILD_DIR); \
+	else \
+		echo 'Makefile not found, use kustomize and kubectl to apply resources.'; \
+		component_build_dir=./$(component_path)/$(BUILD_DIR); \
+		rm -rf $$component_build_dir && mkdir -p $$component_build_dir; \
+		kustomize build --load-restrictor LoadRestrictionsNone -o $$component_build_dir ./$(component_path); \
+		kubectl --context=$(KFCTXT) apply -f $$component_build_dir; \
+	fi
```
Bobgy, line 72: Makefile rules are usually deterministic, they accept no value and produce stable results.
Consider converting this rule to a bash script? I think that simplifies this rule and also make it easier to understand what's going on.
(You may put individual bash scripts at `hack` folder)

kubeflow/Makefile
```diff
@@ -51,6 +52,67 @@ apply-cnrm:
...
+	@for component in $(components) ; do \
+	$(MAKE) apply-component component_path=$$component || break ; \
+	done
+
+	# Kick the IAP pod because we will reset the policy and need to patch it.
+	# TODO(https://github.com/kubeflow/gcp-blueprints/issues/14)
+	kubectl --context=$(KFCTXT) -n istio-system delete pods -l service=iap-enabler
+	# Kick the backend updater pod, because information might be outdated after the apply.
+	# https://github.com/kubeflow/gcp-blueprints/issues/160
+	kubectl --context=$(KFCTXT) -n istio-system delete pods -l service=backend-updater
```
Bobgy, line 114: can these be moved to ingress/IAP package?

kubeflow/common/cert-manager/Makefile
```diff
@@ -0,0 +1,12 @@
+
```
Bobgy, line 1: add a default build_dir, so that the Makefile can run on its own too?

```
build_dir?=./build
```

kubeflow/common/cert-manager/Makefile
```diff
@@ -0,0 +1,12 @@
...
+.PHONY: apply
+apply:
+	# Common cert-manager
+	rm -rf $(build_dir) && mkdir -p $(build_dir)
+	kustomize build --load-restrictor LoadRestrictionsNone -o $(build_dir) ./
```
Bobgy, line 6: nit: I recall that we no longer need load restrictor none, is this case special?

kubeflow/common/cnrm/Makefile
```diff
@@ -29,4 +29,20 @@ wait-gcp:
...
 	   REGION=$(LOCATION) \
-	   NAME=$(NAME) ../../hack/create_context.sh
\ No newline at end of file
+	   NAME=$(NAME) ../../hack/create_context.sh
+
+.PHONY: apply-cnrm
+apply-cnrm:
+ifeq ($(build_dir),)
+	$(error build_dir is not set for apply rule); 
+endif
```
Bobgy, line 38: nit: set a default `build_dir`?

kubeflow/common/iap-secret/Makefile
```diff
@@ -0,0 +1,14 @@
...
+.PHONY: apply
+apply: iap-secret
+
+.PHONY: iap-secret
+iap-secret: check-iap
+	kubectl --context=$(KFCTXT) -n istio-system create secret generic kubeflow-oauth --from-literal=client_id=$(CLIENT_ID) --from-literal=client_secret=$(CLIENT_SECRET) --dry-run=client -o yaml | kubectl apply -f -
+
+.PHONY: check-iap
+check-iap:
+	../../hack/check_oauth_secret.sh
```
Bobgy, line 14: move this script inside?

also, this folder seems to better belong to ingress folder

kubeflow/common/knative/Makefile
```diff
@@ -0,0 +1,23 @@
+# The name of the context for your Kubeflow cluster
+# NAME=$(shell yq r ./Kptfile 'openAPI.definitions."io.k8s.cli.setters.name".x-k8s-cli.setter.value')
+# KFCTXT=$(NAME)
+
+.PHONY: apply
+apply:
+# Common Knative
+# It has error when trying to patch the resource downloaded from knative because it doesn't match kustomize requirement.
+# Original release of knative manifest can not be patched by kustomize because of `already registered id`. For now we run kubectl directly.
```
Bobgy, line 9: this typically means you are combining two packages that have duplicate resources,
a simpler workaround is to figure out what is duplicate and break down apply to multiple steps.

but we can optimize this in a separate PR

kubeflow/common/kubeflow-issuer/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+# Define the self-signed issuer for Kubeflow
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+namespace: cert-manager
+resources:
+- ../cert-manager/cert-manager-1-3/cert-manager/kubeflow-issuer
```
Bobgy, line 6: nit: simply move that folder here?

Bobgy (commented):
/lgtm
with some nit pickings

kubeflow/Makefile
zijianjoy: That makes sense, I had a hard time writing bash script in make rule too. Extracted it to a hack/apply_component.sh.

kubeflow/Makefile
zijianjoy: Moved them to a make rule in common/iap-ingress

kubeflow/common/cert-manager/Makefile
zijianjoy: Make sense, done.

kubeflow/common/cert-manager/Makefile
zijianjoy: Good point, I should clean them up. Done.

kubeflow/common/cnrm/Makefile
zijianjoy: Make sense, done.

kubeflow/common/iap-secret/Makefile
zijianjoy: Sounds good, moved shell script to here. Since I want to apply this iap secret in the middle of the loop that applies components. I make this folder and expose the `make apply` target.

kubeflow/common/knative/Makefile
zijianjoy: Ack, thank you for the info! Currently I am using David Spek's approach to breakdown the duplicated items and adjust to kustomize build rule. We can further optimize it later.

kubeflow/common/kubeflow-issuer/kustomization.yaml
zijianjoy: I am a bit hesitant to move an upstream folder to another folder. It will make future version upgrade a bit complicated. Another solution is to build a Makefile under common/cert-manager which will build both cert-manager and kubeflow-issuer. How do you think?

Bobgy: Adding this to cert-manager make sense, as that's how upstream organizes the folder.

Bobgy: @zijianjoy I am strongly against any references between packages, because goal of this refactor is to make them independent.

zijianjoy: Sounds good, it makes sense to me to establish the independency between packages. I have merged the kubeflow-issuer to cert-manager as in https://github.com/kubeflow/gcp-blueprints/pull/238/files.

PR #235, opened 2021-04-26
(Makefile) Address comments from PR#234

zijianjoy:
To address comments in https://github.com/kubeflow/gcp-blueprints/pull/234.

Please see the response from the above PR link. Thank you for your comment!


kubeflow/common/iap-secret/Makefile
```diff
@@ -11,4 +11,4 @@ iap-secret: check-iap
 
 .PHONY: check-iap
 check-iap:
-	../../hack/check_oauth_secret.sh
+	./check_oauth_secret.sh
```
Bobgy, line 14: Can you clarify why this has to be a separate step than ingress-iap?

kubeflow/Makefile
```diff
@@ -103,15 +83,10 @@ common/kubeflow-issuer \
...
+	BUILD_DIR=$(BUILD_DIR) NAME=$(NAME) KFCTXT=$(KFCTXT) LOCATION=$(LOCATION) PROJECT=$(PROJECT) ./hack/apply_component.sh -path $$component; \
 	done
 
-	# Kick the IAP pod because we will reset the policy and need to patch it.
-	# TODO(https://github.com/kubeflow/gcp-blueprints/issues/14)
-	kubectl --context=$(KFCTXT) -n istio-system delete pods -l service=iap-enabler
-	# Kick the backend updater pod, because information might be outdated after the apply.
-	# https://github.com/kubeflow/gcp-blueprints/issues/160
-	kubectl --context=$(KFCTXT) -n istio-system delete pods -l service=backend-updater
+	KFCTXT=$(KFCTXT) $(MAKE) -C common/iap-ingress pod-reset
```
Bobgy, line 89: Thank you, probably good enough for now!
It will be helpful to identify why it needs to run at the end and which external dependencies it has and try to avoid this.

Bobgy (commented):
/LGTM
With two optional comments

kubeflow/common/iap-secret/Makefile
zijianjoy: After thinking about it, I merged the iap-secret to iap-ingress/ folder. I was wondering whether it makes sense to merge two into one: is there scenario where user wants to only update one folder, but not the other. I think this case is not very common, so sounds good on merging this iap secret to iap-ingress/ folder: https://github.com/kubeflow/gcp-blueprints/pull/238/files

kubeflow/Makefile
zijianjoy: Looks like we need to enforce the backend updater to reconcile for a healthy backend. Maybe we want to wait for iap-ingressgateway to be ready first.

Issue #144, opened 2020-10-29
Upgrade ASM from 1.4 to 1.6+


jlewi:
We are currently using ASM 1.4. Docs currently recommend 1.6.
https://cloud.google.com/service-mesh/docs/gke-install-asm

I don't think 1.4 is still supported.

kubeflow/kubeflow#5313 is tracking ISTIO 1.6 in Kubeflow

Bobgy:
Agree, added to https://github.com/kubeflow/gcp-blueprints/issues/139.
We need to confirm the CRDs we are relying on, like Istio RBAC are still supported on 1.6.

Bobgy:
I think I didn't realize how important this is, the fact that ASM 1.4 is no longer supported make this a high priority for us to migrate.

Bobgy:
Documentation to refer:
https://cloud.google.com/service-mesh/docs/scripted-install/asm-onboarding
https://cloud.google.com/service-mesh/docs/enable-optional-features

Bobgy:
Official doc to refer: https://istio.io/latest/docs/concepts/what-is-istio/

zijianjoy:
Latest istioctl-asm download page: http://cloud/service-mesh/docs/downloading-istioctl#mac-os

zijianjoy:
Attempted to run `make apply` directly after upgrading istioctl-asm. See the following error for command:

command:
`istioctl manifest generate -f $(MANIFESTS_DIR)/gcp/v2/asm/istio-operator.yaml -o $(BUILD_DIR)/istio`

error:

```
#************************************************************************************
# hydrate asm
istioctl manifest generate -f ./upstream/manifests/gcp/v2/asm/istio-operator.yaml -o .build/istio
Error: validation errors (use --force to override):
unknown field "env" in v1alpha1.ProxyConfig
make: *** [hydrate-asm] Error 1
```

zijianjoy:
Exploring applying https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages/blob/master/asm/istio/istio-operator.yaml

zijianjoy:
Making changes in `Makefile` to simulate manual steps in https://cloud.google.com/service-mesh/docs/scripted-install/gke-install.

Currently facing issues:

*  `sni_hosts` are not recognized because validation has been enforced on this field. Changing them to `sniHosts` in `kf-istio-resources.html` fixes the issue.

[...]

PR #237, opened 2021-04-26
(ASM) Add instruction about how to upgrade ASM. Partial #144

zijianjoy:
Partial #144

Validated that upgrading works with ASM-193-2 config 2. 

Note: We might need to educate user to remove the old control plane `istiod-asm-192-1`, once they validate that upgrade is working properly.


kubeflow/common/asm/Makefile
```diff
@@ -3,7 +3,13 @@
...
 # The part before colon symbol should be used for ASM_PACKAGE_VERSION.
 # The part after colon symbol should be used for INSTALL_ASM_SCRIPT_VERSION.
-# You might need to change corresponding kpt value in kubeflow/kpt-set.sh if you upgrade 
+# IMPORTANT: You must change corresponding kpt value in kubeflow/env.sh if you upgrade 
+# For example: export ASM_LABEL=asm-193-2
+# Then go back to kubeflow/ directory, run the following commands:
+# 1. source ./env.sh
+# 2. bash ./kpt-set.sh
+# 3. make apply
+# The above commands allow you to set new namespace labels and re-apply your resources.
```
Bobgy, line 12: This part seems problematic to me, if users need to upgrade asm package, they should be able to do so directly here. Otherwise, they can only apply Kubeflow as a whole, but not parts.

You might need to learn how to set up a sub kpt package in this directory, in that way, you can add kpt setters in Kptfile of this package. hint: refer to https://googlecontainertools.github.io/kpt/reference/pkg/init/

zijianjoy: Thank you for the reference! 

Migrate ASM to a newer version in cluster requires re-applying workload: https://cloud.google.com/service-mesh/docs/scripted-install/gke-upgrade#deploying_and_redeploying_workloads. So we need to restart workload. Let's discuss more about how to do it for other Kubeflow components.

But I think it makes sense in terms of setting a kpt package to read the ASM label value.

zijianjoy: I have updated the ASM upgrade process and local environment variable + kpt config,

kubeflow/kpt-set.sh
```diff
@@ -13,4 +14,4 @@ kpt cfg set -R .  gcloud.compute.zone ${LOCATION}
 kpt cfg set -R .  location ${LOCATION}
 kpt cfg set -R .  log-firewalls false
 
-kpt cfg set -R .  email <YOUR_EMAIL_ADDRESS>
\ No newline at end of file
+kpt cfg set -R .  email <YOUR_EMAIL_ADDRESS>
```
Bobgy, line 17: nit: include ADMIN_EMAIL var in env.sh? then this can be consistent

kubeflow/common/asm/Makefile
```diff
@@ -3,16 +3,22 @@
...
+# 2: Update kubeflow/common/asm/env.sh for the ASM version, this looks like 
+#     asm-192-1, and you can find the exact value in istiod pod's label as well.
+#     For example: You can find label `istio.io/rev: asm-192-1` in istiod-asm-192-1
+#     service. Then run `source env.sh` under `common/asm/` folder.
+# 3: Run `bash kpt-set.sh` under `common/asm` folder. It will configure kpt setter
+#     value for all components which requires to set ASM namespace label.
+# 4: Follow official doc: https://cloud.google.com/service-mesh/docs/scripted-install/gke-upgrade#deploying_and_redeploying_workloads
+#    to migrate workloads in each namespace, including user namespace created
+#    in https://www.kubeflow.org/docs/components/multi-tenancy/getting-started/.
+# 5: Alternatively, hydrate all Kubeflow resources to keep track of latest changes.
```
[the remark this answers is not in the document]
zijianjoy: Since we advice user to follow official ASM doc in step 4, we don't need to **apply** all Kubeflow resources.

However, this step allows users to **hydrate** `build/` result, so the resource yaml file has the new ASM labels. Because it is possible that we add or remove namespaces which requires ASM label, but we might not remember to come back here to update this comment, so I didn't list the specific resources that need changes here.

kubeflow/common/asm/README.md
```diff
@@ -0,0 +1,29 @@
+# .
```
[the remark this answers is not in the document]
zijianjoy: I will remove this file.

kubeflow/kpt-set.sh
zijianjoy: This is a very good point! I will make the corresponding changes.

kubeflow/env.sh
```diff
@@ -5,6 +5,4 @@ export MGMT_NAME=<management-cluster-name>
 export MGMTCTXT=${MGMT_NAME}
 export LOCATION=<zone>
 
-export ASM_LABEL=asm-192-1
-
-
+source ./common/asm/env.sh
```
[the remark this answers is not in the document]
zijianjoy: We are already using ASM specific setup: For example, profile controller will remove the sidecar injection label and use ASM label. And the existing `kubeflow/env.sh` is defining ASM_LABEL env variable currently. 

So if users want to replace asm with istio, it requires many steps for them already, and this installation approach is not validated by us. (Istio/ASM is the foundation of many things: IAP, gateway, ingress, sidecar, etc.)

I want to make sure we only define the ASM label once in the whole directory, this is why I call the `common/asm/env.sh` script from root `env.sh`. And asking user to run `kpt cfg set -R asm-label ...` as single installation step by default seems to be redundant since we already ask them to run `bash kpt-set.sh` under `kubeflow/` directory. And keeping all shell commands in a bash script allows us to track it using git. Does it make sense to you?

kubeflow/common/asm/Makefile
```diff
@@ -3,16 +3,22 @@
...
+# ASM Upgrade Guide
+# 1: Determine the ASM version you want to upgrade, follow the instruction above
+#     to update ASM_PACKAGE_VERSION and INSTALL_ASM_SCRIPT_VERSION. Then run
+#     `make install-asm` to install new ASM version.
+# 2: Update kubeflow/common/asm/env.sh for the ASM version, this looks like 
+#     asm-192-1, and you can find the exact value in istiod pod's label as well.
+#     For example: You can find label `istio.io/rev: asm-192-1` in istiod-asm-192-1
+#     service. Then run `source env.sh` under `common/asm/` folder.
+# 3: Run `bash kpt-set.sh` under `common/asm` folder. It will configure kpt setter
+#     value for all components which requires to set ASM namespace label.
```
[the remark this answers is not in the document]
zijianjoy: I agree that it can be simplified to one line of shell call. The reason I write this in a shell script is that:

1. The root env.sh and kpt-set.sh will need them for full fledged deployment
2. Keeping track of kpt setter and environment variables in version control style (git). 

I think education of `kpt` will need its own section in Kubeflow Google Cloud instruction. The full fledged Kubeflow deployment guide serves a single purpose: Allows user to deploy a working solution of Kubeflow. Once they have a cluster ready, it will make their `kpt` learning much easier since they can compare the effect of their kpt commands based on a working deployment. Does it make sense?

kubeflow/kpt-set.sh
```diff
@@ -1,16 +1,17 @@
...
-kpt cfg set -R .  mgmt-ctxt ${MGMT_NAME}
+kpt cfg set -R .  mgmt-ctxt "${MGMT_NAME}"
 
-kpt cfg set -R .  name ${KF_NAME}
-kpt cfg set -R .  gcloud.project.projectNumber ${KF_PROJECT_NUMBER}
-kpt cfg set -R .  gcloud.core.project ${KF_PROJECT}
-kpt cfg set -R .  gcloud.compute.zone ${LOCATION}
-kpt cfg set -R .  location ${LOCATION}
+kpt cfg set -R .  name "${KF_NAME}"
+kpt cfg set -R .  gcloud.project.projectNumber "${KF_PROJECT_NUMBER}"
```
Bobgy, line 11: move this to last line

kubeflow/common/asm/env.sh
```diff
@@ -0,0 +1,2 @@
+
+export ASM_LABEL=asm-192-1
```
Bobgy, line 2: move to folder root

kubeflow/env.sh
```diff
@@ -1,12 +1,10 @@
 export KF_NAME=<kubeflow-cluster-name>
 export KF_PROJECT=<gcp-project-id>
-export KF_PROJECT_NUMBER=$(gcloud projects describe ${KF_PROJECT} --format='value(projectNumber)')
+export KF_PROJECT_NUMBER=$(gcloud projects describe "${KF_PROJECT}" --format='value(projectNumber)')
```
zijianjoy, line 3: Move it to bottom to avoid potential failure issue

kubeflow/kpt-set.sh
zijianjoy: Sounds good, moved to the bottom.

Bobgy (approved):
/lgtm
/approve

Bobgy:
BTW, the best would be upgrading ASM to latest before release.

PR #238, opened 2021-04-27
(Makefile) Integrate iap-secret, kubeflow-issuer to their own parent packages

zijianjoy:
Adjusting comments from https://github.com/kubeflow/gcp-blueprints/pull/234 and https://github.com/kubeflow/gcp-blueprints/pull/235


Bobgy:
Thanks! LGTM

PR #241, opened 2021-04-28
(README) Update deployment instruction for Kubeflow and Management clusters

zijianjoy:
Part of #230

Adding README as the instruction of deploying Kubeflow. We should update README accordingly when making changes to deployment steps.


README.md
```diff
@@ -1,15 +1,19 @@
-# Kubeflow Blueprint
+# Kubeflow Blueprint On GCP (Google Cloud Platform)
```
Bobgy, line 1: Note, the official branding for GCP is now `Google Cloud`, we should avoid `GCP` or `Google Cloud Platform` in user facing documentation (it's fine here, but would be nicer if we stick to the convention from the start).

Bobgy (commented):
/lgtm
/approve
only one nit picking

PR #242, opened 2021-04-28
chore: clean up common/iap-ingress, rm unused packages

Bobgy:
/assign @zijianjoy


kubeflow/common/iap-ingress/Makefile
```diff
@@ -8,7 +8,7 @@ apply: iap-secret
 
 .PHONY: iap-secret
 iap-secret: check-iap
-	kubectl --context=$(KFCTXT) -n istio-system create secret generic kubeflow-oauth --from-literal=client_id=$(CLIENT_ID) --from-literal=client_secret=$(CLIENT_SECRET) --dry-run=client -o yaml | kubectl apply -f -
+	kubectl --context=$(KFCTXT) -n istio-system create secret generic kubeflow-oauth --from-literal=client_id=$(CLIENT_ID) --from-literal=client_secret=$(CLIENT_SECRET) --dry-run -o yaml | kubectl apply -f -
```
Bobgy, line 11: this can be a little controversial, how can we be compatible with latest kubectl?
I think we should assume users use kubectl of version around 1.17 because kubectl and kubernetes version compatibility is within 2 minor version diff, see https://kubernetes.io/docs/setup/release/version-skew-policy/#kubectl.

So I think we can just make this change, WDUT?

kubeflow/common/iap-ingress/base/cert.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: networking.gke.io/v1beta1
+apiVersion: networking.gke.io/v1
```
Bobgy, line 1: note, v1beta1 is deprecated, upgrade to v1 schema -- but the fields we used are compatible

kubeflow/common/iap-ingress/Makefile
zijianjoy: I think it is okay to apply this change. Thank you Yuan!

kubeflow/common/iap-ingress/base/cert.yaml
zijianjoy: Thank you Yuan! Sounds good.

zijianjoy (commented):
/lgtm Thank you Yuan for the changes! I only have one comment.

kubeflow/common/iap-ingress/base/kustomization.yaml
```diff
@@ -27,82 +27,56 @@ images:
...
   objref:
     kind: ConfigMap
-    name: parameters
+    name: iap-ingress-config
     apiVersion: v1
   fieldref:
     fieldpath: data.istioNamespace
 configurations:
-- params.yaml
+- ../base/params.yaml
```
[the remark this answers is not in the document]
Bobgy: Ohh, you are right!

PR #245, opened 2021-04-28
(iap-ingress) Patch BackendConfig healthcheck port with the correct port number

fabito:
# Problem
Unhealthy backend due to misconfigured health check.
The `iap-backendconfig` BackendConfig healthCheck spec points to the wrong port. It overwrites `backend-updater` changes.

# Solution
Refactor `update_backend.sh` script (used by `backend-updater` stateful set) to patch the BackendConfig `iap-backendconfig` healthcheck port instead of updating the health-check resource directly.

### DIFF
--- a/kubeflow/common/iap-ingress/base/cluster-role.yaml
+++ b/kubeflow/common/iap-ingress/base/cluster-role.yaml
@@ -17,8 +17,10 @@ rules:
 - apiGroups:
   - extensions
   - networking.k8s.io
+  - apps
   resources:
   - ingresses
+  - deployments
   verbs:
   - get
   - list
--- a/kubeflow/common/iap-ingress/base/cluster-role.yaml
+++ b/kubeflow/common/iap-ingress/base/cluster-role.yaml
@@ -39,3 +39,12 @@ rules:
   - virtualservices
   verbs:
   - '*'
+- apiGroups:
+  - cloud.google.com
+  resources:
+  - backendconfigs
+  verbs:
+  - get
+  - list
+  - patch
+  - update
--- a/kubeflow/common/iap-ingress/base/config-map.yaml
+++ b/kubeflow/common/iap-ingress/base/config-map.yaml
@@ -127,13 +127,12 @@ data:
     [ -z ${NAMESPACE} ] && echo Error NAMESPACE must be set && exit 1
 
     set_health_check () {
-      # We need the nodeport for istio-ingressgateway status-port
       # See: https://cloud.google.com/service-mesh/docs/iap-integration#deploying_the_load_balancer
       HC_INGRESS_PORT=$(kubectl -n istio-system get service istio-ingressgateway -o jsonpath='{.spec.ports[?(@.name=="status-port")].nodePort}')
       HC_INGRESS_PATH=$(kubectl -n istio-system get deployments istio-ingressgateway -o jsonpath='{.spec.template.spec.containers[?(@.name=="istio-proxy")].readinessProbe.httpGet.path}')
 
       echo Setting BackendConfig healthCheck.port to: ${HC_INGRESS_PORT} and healthCheck.requestPath to ${HC_INGRESS_PATH}
-      kubectl patch iap-backendconfig -n ${NAMESPACE} --type json -p '[{"op": "replace", "path": "/spec/healthCheck/port", "value": '${HC_INGRESS_PORT}'}, {"op": "replace", "path": "/spec/healthCheck/requestPath", "value": "${HC_INGRESS_PATH}"}]'
+      kubectl patch backendconfig iap-backendconfig -n ${NAMESPACE} --type json -p '[{"op": "replace", "path": "/spec/healthCheck/port", "value": '${HC_INGRESS_PORT}'}, {"op": "replace", "path": "/spec/healthCheck/requestPath", "value": "'${HC_INGRESS_PATH}'"}]'
     }
 
     while true; do
--- a/kubeflow/common/iap-ingress/base/config-map.yaml
+++ b/kubeflow/common/iap-ingress/base/config-map.yaml
@@ -125,8 +125,49 @@ data:
     set -x
 
     [ -z ${NAMESPACE} ] && echo Error NAMESPACE must be set && exit 1
+    [ -z ${SERVICE} ] && echo Error SERVICE must be set && exit 1
+    [ -z ${INGRESS_NAME} ] && echo Error INGRESS_NAME must be set && exit 1
+
+    PROJECT=$(curl -s -H "Metadata-Flavor: Google" http://metadata.google.internal/computeMetadata/v1/project/project-id)
+    if [ -z ${PROJECT} ]; then
+      echo Error unable to fetch PROJECT from compute metadata
+      exit 1
+    fi
+    
+    if [[ ! -z "${GOOGLE_APPLICATION_CREDENTIALS}" ]]; then
+      # TODO(jlewi): As of 0.7 we should always be using workload identity. We can remove it post 0.7.0 once we have workload identity
+      # fully working
+      # Activate the service account, allow 5 retries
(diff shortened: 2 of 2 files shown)


Bobgy:
/assign @zijianjoy 
to review

kubeflow/common/iap-ingress/base/config-map.yaml
```diff
@@ -178,11 +178,11 @@ data:
         gcloud --project=${PROJECT} compute health-checks update http ${HEALTH_CHECK_URI} --request-path=${HEALTHCHECK_PATH}
```
zijianjoy, line 178: Should we change this to patch `iap-backendconfig` with ${HEALTHCHECK_PATH}?

kubeflow/common/iap-ingress/base/config-map.yaml
```diff
@@ -178,11 +178,11 @@ data:
...
       else
         # /healthz/ready is the health check path for istio-ingressgateway
-        echo Running health checks update ${HEALTH_CHECK_URI} with /healthz/ready
-        gcloud --project=${PROJECT} compute health-checks update http ${HEALTH_CHECK_URI} --request-path=/healthz/ready
```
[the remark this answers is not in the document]
fabito: Yep, makes total sense.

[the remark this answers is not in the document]
fabito: I thought about that.
I wanted this script to change the minimum possible. I think that the blueprint manifests should be the source of truth.
We only need to patch the port because it is randomly assigned, ideally we would set it to a fixed number.

I think that the health check resource is actually created from the `BackendConfig`.
When we fetch `HEALTH_CHECK_URI` :

```bash
HEALTH_CHECK_URI=$(gcloud compute --project=${PROJECT} health-checks list --filter=name~${BACKEND_NAME} --uri)
```
Aren't we just fetching the same value we defined in the yaml ?

kubeflow/common/iap-ingress/base/config-map.yaml
fabito: I am not sure. I haven't changed that flow cause I didn't know exactly its purpose.
If it is just a customisation point, shouldn't we just apply a kustomize overlay/patch (to change `BackendConfig`) ?

zijianjoy (commented):
Thank you @fabito for the change!

I think it makes sense to try updating BackendConfig to match the health check port. However, it will also mean that the built resource in local repo doesn't match the actual deployment value. 

Maybe we can explore `kpt` setter: Read the setter value from https://github.com/kubeflow/gcp-blueprints/blob/master/kubeflow/common/asm/asm/istio/options/iap-operator.yaml#L31, and set it in https://github.com/kubeflow/gcp-blueprints/blob/master/kubeflow/common/iap-ingress/base/backend-config.yaml#L13. But it might go beyond the scope of this PR, so feel free to focus on backend updater here.

kubeflow/common/iap-ingress/base/config-map.yaml
fabito: Agree, looks like its obsolete now, I will remove.

kubeflow/common/iap-ingress/base/config-map.yaml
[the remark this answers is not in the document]
fabito: Hi @zijianjoy ,

I am still confused.
Isn't the health check created and configured based on the `BackendConfig`'s spec.healthCheck definition ?

```yaml
apiVersion: cloud.google.com/v1beta1
kind: BackendConfig
metadata:
  name: iap-backendconfig
spec:
  healthCheck:
    port: 31223
    type: HTTP
    requestPath: /healthz/ready # <===
```
I `requestPath` is already set to `/healthz/ready` in the yaml why would we need to reset it using gcloud ?
Am I missing something ?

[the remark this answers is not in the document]
fabito: In included an patch op for the requestPath

kubeflow/common/iap-ingress/base/cluster-role.yaml
```diff
@@ -17,8 +17,10 @@ rules:
 - apiGroups:
   - extensions
   - networking.k8s.io
+  - apps
   resources:
   - ingresses
+  - deployments
```
zijianjoy, line 23: Why do we need to add `apps` and `deployments` permission?

kubeflow/common/iap-ingress/base/config-map.yaml
```diff
@@ -125,65 +125,14 @@ data:
...
-      while [[ -z ${HEALTH_CHECK_URI} ]];
-      do HEALTH_CHECK_URI=$(gcloud compute --project=${PROJECT} health-checks list --filter=name~${BACKEND_NAME} --uri);
-      echo "Waiting for the healthcheck resource PROJECT=${PROJECT} NODEPORT=${NODE_PORT} SERVICE=${SERVICE}...";
-      sleep 2;
-      done
+      # See: https://cloud.google.com/service-mesh/docs/iap-integration#deploying_the_load_balancer
+      HC_INGRESS_PORT=$(kubectl -n istio-system get service istio-ingressgateway -o jsonpath='{.spec.ports[?(@.name=="status-port")].nodePort}')
+      HC_INGRESS_PATH=$(kubectl -n istio-system get deployments istio-ingressgateway -o jsonpath='{.spec.template.spec.containers[?(@.name=="istio-proxy")].readinessProbe.httpGet.path}')
 
-      echo health check URI is ${HEALTH_CHECK_URI}
```
zijianjoy, line 168: We still need these parts because it makes sure we check for backend existence. Would you like to revert the changes so we can focus our scope in current PR? We can discuss more in a separate issue. I only suggest removing the block within `if [[ ${HEALTHCHECK_PATH} ]]; then`

kubeflow/common/iap-ingress/base/config-map.yaml
```diff
@@ -125,65 +125,14 @@ data:
...
-      else
-        # /healthz/ready is the health check path for istio-ingressgateway
-        echo Running health checks update ${HEALTH_CHECK_URI} with /healthz/ready
-        gcloud --project=${PROJECT} compute health-checks update http ${HEALTH_CHECK_URI} --request-path=/healthz/ready
-        # We need the nodeport for istio-ingressgateway status-port
-        STATUS_NODE_PORT=$(kubectl -n istio-system get service istio-ingressgateway -o jsonpath='{.spec.ports[?(@.name=="status-port")].nodePort}')
-        gcloud --project=${PROJECT} compute health-checks update http ${HEALTH_CHECK_URI} --port=${STATUS_NODE_PORT}
-      fi      
+      echo Setting BackendConfig healthCheck.port to: ${HC_INGRESS_PORT} and healthCheck.requestPath to ${HC_INGRESS_PATH}
+      kubectl patch backendconfig iap-backendconfig -n ${NAMESPACE} --type json -p '[{"op": "replace", "path": "/spec/healthCheck/port", "value": '${HC_INGRESS_PORT}'}, {"op": "replace", "path": "/spec/healthCheck/requestPath", "value": "'${HC_INGRESS_PATH}'"}]'
```
zijianjoy, line 135: NIT: Can we break down the patch into two commands?

```
kubectl patch backendconfig iap-backendconfig -n ${NAMESPACE} --type json -p '[{"op": "replace", "path": "/spec/healthCheck/port", "value": '${HC_INGRESS_PORT}'}]'
kubectl patch backendconfig iap-backendconfig -n ${NAMESPACE} --type json -p '[{"op": "replace", "path": "/spec/healthCheck/requestPath", "value": "'${HC_INGRESS_PATH}'"}]'
```

This is for easier debugging by separating individual patches, and for readability on the backend updater script.

kubeflow/common/iap-ingress/base/cluster-role.yaml
fabito: That's for fetching the `requestPath`

zijianjoy (approved):
/lgtm /approve @fabito Thank you so much for contributing to this PR!

PR #247, opened 2021-04-29
(Makefile) Finish complete lifecycle of kubeflow management, clean up old commands. Fix #230

### LINKED ISSUE #230: Simplify Makefile

zijianjoy:
Fix #230

Implement following commands:

`make hydrate-all`
`make apply-all`
`make clean-build`
`make delete-gcp`

Make sure other items align with these commands and deploy successfully.


kubeflow/common/istio/kustomization.yaml
```diff
@@ -1,4 +1,4 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
-- ./upstream/istio-1-9-0/kubeflow-istio-resources/base # {"$kpt-set":"istio-istio-base"}
```
zijianjoy, line 4: It looks like `istio-1-9-0/` is not created when pulling from upstream: `bash ./pull-upstream.sh`

Bobgy (commented):
/lgtm
/approve
This is awesome! Love how this simplified the global Makefile significantly
we can address comments separately

Issue #254, opened 2021-05-03
set-values.sh should include an env var for ADMIN_EMAIL


zijianjoy:
Also, shall we keep kpt cfg set commands in documentation? If we prepare a script of set-values, users will think it's an abstraction, but if we let them use commands directly, that lets them learn kpt quicker.

example:
https://www.kubeflow.org/docs/distributions/gke/deploy/management-setup/

PR #255, opened 2021-05-03
(env.sh) Move all envrionment variable config to env.sh. Partial #254.

zijianjoy:
Partial #254.

Feel free to comment. I am going to merge it now.

### DIFF
--- a/kubeflow/env.sh
+++ b/kubeflow/env.sh
@@ -1,9 +1,11 @@
 export KF_NAME=<kubeflow-cluster-name>
 export KF_PROJECT=<gcp-project-id>
+export KF_PROJECT_NUMBER=$(gcloud projects describe ${KF_PROJECT} --format='value(projectNumber)')
 export KF_DIR=<kubeflow-download-path>
 export MGMT_NAME=<management-cluster-name>
 export MGMTCTXT=${MGMT_NAME}
 export LOCATION=<zone>
+export ADMIN_EMAIL=<administrator-full-email-address>
 
 export ASM_LABEL=asm-192-1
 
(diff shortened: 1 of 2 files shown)


kubeflow/env.sh
```diff
@@ -1,9 +1,11 @@
 export KF_NAME=<kubeflow-cluster-name>
 export KF_PROJECT=<gcp-project-id>
+export KF_PROJECT_NUMBER=$(gcloud projects describe ${KF_PROJECT} --format='value(projectNumber)')
```
Bobgy, line 3: Note, if sth goes wrong in $(), the script will not fail, so that's a silent failure and really hard to debug later.

kubeflow/kpt-set.sh
```diff
@@ -7,10 +7,10 @@ kpt cfg set -R .  asm-label ${ASM_LABEL}
 kpt cfg set -R .  mgmt-ctxt ${MGMT_NAME}
 
 kpt cfg set -R .  name ${KF_NAME}
-kpt cfg set -R .  gcloud.project.projectNumber <KF_PROJECT_NUMBER>
+kpt cfg set -R .  gcloud.project.projectNumber ${KF_PROJECT_NUMBER}
```
Bobgy, line 10: nit: always surround variable with ""

E.g. "${KF_PROJECT_NUMBER}", because there can be spaces that make the script fail in unexpected ways

kubeflow/env.sh
zijianjoy: That is a good point, but I tried to fail the script inside $() on purpose, this will stop and output the error.

Do you mean the situation when the command in $() was wrong in producing the output, but the syntax is accurate? Should we advice user to provide project explicitly, instead of programming it for user?

kubeflow/kpt-set.sh
zijianjoy: It is a very good point, thank you for mentioning! I will add them in shell scripts. 

For now I will patch them in https://github.com/kubeflow/gcp-blueprints/pull/237/ for convenient.

PR #262, opened 2021-05-06
(profile-controller) Adopt v1.3.0-gcp profile controller 

zijianjoy:
This is our current way to adopting https://github.com/kubeflow/kubeflow/pull/5761/files.

I am able to create profile and add contributor via UI and yaml file.

Note that even though I have added sidecar injection label ` istio.io/rev=asm-192-2` to kubeflow namespace. I am not able to inject sidecar for KFP yet. It might not be related to this PR.


Bobgy (commented):
/lgtm
/approve

PR #272, opened 2021-05-12
feat: refactor KFP managed storage a separate package. Fixes #275

### LINKED ISSUE #275: Separate cloud-sql, bucket from KFP, document its behavior, update Makefile.

Bobgy:
Fixes #275


Bobgy:
/assign @zijianjoy 
I haven't troubleshooted the setup, creating the PR for early review.

kubeflow/env.sh
```diff
@@ -1,12 +1,39 @@
...
-export ASM_LABEL=asm-192-1
+# The following env vars are also configurable.
 
+# KF_NAME is your Kubeflow cluster name.
+# It should satisfy the following conditions:
+# * is unique within your project.
+# * alphanumeric characters and hyphen "-" only.
+# * starts with a letter.
+# * ends with an alphanumeric character.
+export KF_NAME=kubeflow
```
zijianjoy, line 27: Should we pre-set this value for user? This variable is very frequently used by user to identify and maintain a cluster, so I think user should explicitly set them: KF_NAME=<Kubeflow-cluster-name>

kubeflow/kpt-set.sh
```diff
@@ -1,22 +1,58 @@
-
-# Use kpt to set kustomize values
+# Set kpt setters using env vars.
 
 set -ex
 
-kpt cfg set -R .  gke.private false
-kpt cfg set -R .  asm-label ${ASM_LABEL}
+if [[ -z "${KF_NAME}" ]]; then
+    echo "KF_NAME env var is required"
```
zijianjoy, line 6: Should we `exit 1` when user doesn't set required env variable?

kubeflow/kpt-set.sh
```diff
@@ -1,22 +1,58 @@
...
+# The CloudSQL instance and Cloud Storage bucket instance are created during
+# deployment, so you should make sure their names are not used before.
+CLOUDSQL_NAME="${KF_NAME}-kfp"
+BUCKET_NAME="${KF_PROJECT}-kfp"
+# common/managed-storage deploys specified CloudSQL and Cloud Storage bucket.
+kpt cfg set common/managed-storage cloudsql-name "${CLOUDSQL_NAME}"
+kpt cfg set common/managed-storage bucket-name "${BUCKET_NAME}"
+# apps/pipelines uses specified CloudSQL and Cloud Storage bucket.
+kpt cfg set apps/pipelines cloudsql-name "${CLOUDSQL_NAME}"
+kpt cfg set apps/pipelines bucket-name "${BUCKET_NAME}"
```
zijianjoy, line 58: NIT: Note that we currently paste the whole content of `env.sh` and `kpt-set.sh` in https://www.kubeflow.org/docs/distributions/gke/deploy/deploy-cli/#environment-variables.

I can foresee that it becomes harder and harder to keep shell script, README and kubeflow.org documentation in sync when this file grows. I recommend using this file as the source of truth by documentation, to prevent out-of-sync.

Makefile
```diff
@@ -1,17 +1,19 @@
 .PHONY: reset-cfg-values
 reset-cfg-values:
-	kpt cfg set -R . gke.private false
+	kpt cfg set -R kubeflow name KUBEFLOW-NAME
 
+	kpt cfg set -R management name MANAGEMENT-NAME
```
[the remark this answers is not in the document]
Bobgy: Agree, I'd also want to rename them when looking at this

kubeflow/README.md
```diff
@@ -62,7 +62,6 @@ Provide actual value for the following variables in `env.sh`:
 ```
 KF_NAME=<kubeflow-cluster-name>
 KF_PROJECT=<gcp-project-id>
-KF_DIR=<current-kubeflow-directory-path>
```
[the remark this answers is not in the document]
Bobgy: KF_DIR is a bash only env var, so I think we can only keep it in documentation. It's kind of problematic in env.sh, because we might want to run env.sh from different working directories.

kubeflow/env.sh
```diff
@@ -1,12 +1,39 @@
...
 
-export ASM_LABEL=asm-192-1
+# The following env vars are also configurable.
 
+# KF_NAME is your Kubeflow cluster name.
+# It should satisfy the following conditions:
+# * is unique within your project.
+# * alphanumeric characters and hyphen "-" only.
+# * starts with a letter.
+# * ends with an alphanumeric character.
```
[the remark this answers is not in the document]
Bobgy: Yes, I don't remember the limit off the top of my head. We can add it later (non-release blocker)

kubeflow/env.sh
Bobgy: I meant, they can always change it as documented above. So provide a valid default value sounds better than an invalid one, what do you think?

kubeflow/kpt-set.sh
Bobgy: Good point, forgot that!

kubeflow/kpt-set.sh
Bobgy: I think that's a reasonable choice

Makefile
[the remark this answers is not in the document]
Bobgy: I'll have to do a separate PR though, because this will cause many changes.

Bobgy:
/assign @zijianjoy 
I revamped according to comments and verified the deployment works e2e now.

Makefile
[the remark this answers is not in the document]
zijianjoy: Yes, we should do it in a separate PR. It is not a blocking issue for release.

kubeflow/README.md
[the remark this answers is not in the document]
zijianjoy: I see, then we should probably extract this line in documentation, and ask user to run manually.

kubeflow/env.sh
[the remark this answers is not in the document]
zijianjoy: If I remember correctly, it is `no more than 25 characters`, I think it is a calculated result because this name is used as a prefix for longer variables. But I don't know the actual formula for getting number 25.

kubeflow/env.sh
zijianjoy: It sounds good for providing a starting point for new users.

kubeflow/kpt-set.sh
zijianjoy: Thank you Yuan for confirming! I will update documentation accordingly

kubeflow/hack/delete_gcp.sh
```diff
@@ -1,9 +1,15 @@
...
-read REPLY; 
-if [[ $REPLY =~ ^[Yy]$ ]]
-then 
-    kubectl --context=$MGMTCTXT delete -f common/cnrm/$BUILD_DIR
+if [[ -z "${MGMTCTXT}" ]]; then
+    echo "MGMTCTXT env var is required"
+    exit 1
+fi
+
+echo "Deleting all GCP resources will cause destruction of all services and data on this cluster. Confirm? [y/N]";
```
zijianjoy, line 9: NIT:`... will cause destruction of all services and data except Cloud SQL instance and GCS bucket. ```

zijianjoy (commented):
/lgtm
/approve

Thank you Yuan for the awesome work and detailed documentation!  Only a few NITs, merging this PR and we can follow up separately for comments.

Makefile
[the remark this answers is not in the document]
Bobgy: I tried, but it's a little hard -- also I start to feel that setting a name for kubeflow/mangement folder seems like a good enough abstraction, WDUT?

PR #274, opened 2021-05-13
feat: allow choosing components. Part of #275

Bobgy:
Part of #275

Change: read components list from config.yaml
Note, this assumes https://github.com/kubeflow/gcp-blueprints/pull/272 is merged first. I've verified deploying a new KF cluster with both https://github.com/kubeflow/gcp-blueprints/pull/272 and #274 and it's working properly.

/assign @zijianjoy


kubeflow/config.yaml
```diff
@@ -0,0 +1,121 @@
...
+# Each component may have a comment like: `dependencies: [ xxx ]`
+# It means this component depends on listed components.
+# Note, that the information may not be fully accurate at the moment. Welcome
+# feedback if you find any mistakes.
+
+# You can delete/comment lines to selectively pick the components you want to
+# deploy.
+# If you want to customize one component, edit or overlay manifests of that
+# component.
+components:
```
zijianjoy, line 13: It is a very good approach for component customization!

Should we call out that `the order of component list` here matters? Since some components depends on others, it will cause error if we swap the component order here. For example: cnrm should be deployed before any applciations.

zijianjoy:
/lgtm
/approve

Thank you Yuan! I like the components listed in yaml file, and the dependency explanation.

PR #278, opened 2021-05-14
chore: delete unused packages folder

Bobgy:
/assign @zijianjoy


Bobgy:
I think this is very straight forward, so merging first.

Although some packages removed might still be useful in the future, we can always recover their content from this PR in the history.

PR #279, opened 2021-05-14
chore: clean up doc + make Makefiles usable individually

Bobgy:
/assign @zijianjoy


Bobgy:
I was wrong, mesh_id is actually required

Bobgy:
OK, now verified deployment works properly.

kubeflow/apps/kfserving/Makefile
```diff
@@ -1,4 +1,7 @@
 build_dir?=./build
+NAME=$(shell yq r ./Kptfile 'openAPI.definitions."io.k8s.cli.setters.name".x-k8s-cli.setter.value')
+# The kubectl context for your Kubeflow cluster
+KFCTXT=$(NAME)
```
zijianjoy, line 4: We also need to define $(PROJECT) because it is being used in `make apply`

kubeflow/hack/delete_gcp.sh
```diff
@@ -1,12 +1,12 @@
...
+set -e
 
 if [[ -z "${MGMTCTXT}" ]]; then
     echo "MGMTCTXT env var is required"
     exit 1
 fi
 
-echo "Deleting all GCP resources will cause destruction of all services and data on this cluster. Confirm? [y/N]";
+echo "Deleting all Google Cloud resources including your GKE cluster and data in the cluster, except the Cloud SQL instance and GCS bucket. Confirm? [y/N]";
 read REPLY;
```
zijianjoy, line 10: I think we can merge these two lines by:

read -p "Deleting all ..." REPLY;

I will make this change from my side.

zijianjoy (commented):
/lgtm
/approve

Thank you so much Yuan for the clean up! I will follow up on the comments I made, merging this change.

PR #284, opened 2021-05-28
feat: make KFP v2 compatible mode work out of the box

Bobgy:
Part of https://github.com/kubeflow/pipelines/issues/5680

Depends on https://github.com/kubeflow/pipelines/pull/5750

Changes:
* configure default pipeline root to gcs bucket using a new kpt setter `default-pipeline-root`
* extract KFP specific pull-upstream script to its own folder, so that admin can only update KFP and deploy

### DIFF
--- a/kubeflow/apps/pipelines/pull-upstream.sh
+++ b/kubeflow/apps/pipelines/pull-upstream.sh
@@ -1,0 +1,31 @@
+#!/usr/bin/env bash
+#
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -ex
+
+# export KUBEFLOW_PIPELINES_VERSION=1.5.0
+# export KUBEFLOW_PIPELINES_REPO=https://github.com/kubeflow/pipelines.git
+export KUBEFLOW_PIPELINES_VERSION=config-default-pipeline-root
+export KUBEFLOW_PIPELINES_REPO=https://github.com/Bobgy/pipelines.git
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" > /dev/null && pwd)"
+
+cd "${DIR}"
+if [ -d upstream ]; then
+    rm -rf upstream
+fi
+mkdir -p upstream
+kpt pkg get "${KUBEFLOW_PIPELINES_REPO}/manifests/kustomize/@${KUBEFLOW_PIPELINES_VERSION}" upstream
+mv upstream/kustomize/* upstream
--- a/kubeflow/apps/pipelines/pull-upstream.sh
+++ b/kubeflow/apps/pipelines/pull-upstream.sh
@@ -16,6 +16,7 @@
 
 set -ex
 
+# TODO(Bobgy): use KFP 1.6.1 when https://github.com/kubeflow/pipelines/pull/5750 is released.
 # export KUBEFLOW_PIPELINES_VERSION=1.5.0
 # export KUBEFLOW_PIPELINES_REPO=https://github.com/kubeflow/pipelines.git
 export KUBEFLOW_PIPELINES_VERSION=config-default-pipeline-root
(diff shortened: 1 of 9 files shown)


Bobgy:
/hold
wait for upstream PR to merge first

kubeflow/pull-upstream.sh
```diff
@@ -1,12 +1,26 @@
...
 
 set -ex
 
 export KUBEFLOW_MANIFESTS_VERSION=v1.3.0
 export KUBEFLOW_MANIFESTS_REPO=https://github.com/kubeflow/manifests.git
 
-export KUBEFLOW_PIPELINES_VERSION=1.5.0
-export KUBEFLOW_PIPELINES_REPO=https://github.com/kubeflow/pipelines.git
+# Pull Kubeflow Pipelines upstream manifests.
+./apps/pipelines/pull-upstream.sh
```
zijianjoy, line 23: Should we use `make -C apps/pipelines pull-upstream` instead?

kubeflow/kpt-set.sh
```diff
@@ -67,3 +67,4 @@ kpt cfg set common/managed-storage bucket-name "${BUCKET_NAME}"
 # apps/pipelines uses specified CloudSQL and Cloud Storage bucket.
 kpt cfg set apps/pipelines cloudsql-name "${CLOUDSQL_NAME}"
 kpt cfg set apps/pipelines bucket-name "${BUCKET_NAME}"
+kpt cfg set apps/pipelines default-pipeline-root "gs://${BUCKET_NAME}/v2/artifacts"
```
zijianjoy, line 70: Should we define `default-pipeline-root` env var in env.sh? So user can set default pipeline root if they want. Or define `default-pipeline-root-path` which is defaulted to `v2/artifacts`. Then we can compose the `default-pipeline-root` using "gs://${BUCKET_NAME}/${default-pipeline-root-path}".

On the other hand, maybe you pin this path on purpose, happy to hear more about rational.

kubeflow/apps/pipelines/pull-upstream.sh
```diff
@@ -0,0 +1,32 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -ex
+
+# TODO(Bobgy): use KFP 1.6.1 when https://github.com/kubeflow/pipelines/pull/5750 is released.
```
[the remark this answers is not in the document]
Bobgy: Our goal is -- each sub folder can be used on its own. Therefore, I think it's better separating pull upstream scripts.

Note, there is feature planned to natively support sub package is kpt. When kpt 1.0 is released we can use the standard mechanism.

kubeflow/kpt-set.sh
Bobgy: My major reason is that this config is not very useful. I expect most people can use the default value. Therefore, I preferred omitting it from env vars.

kubeflow/pull-upstream.sh
Bobgy: This is a good point. I am not sure. If we add it to makefile the model is more consistent with other commands.

If use separate pull-upstream.sh, packages without makefile can easily add them too.

kubeflow/kpt-set.sh
zijianjoy: Sounds good to me for preferring default value.

kubeflow/pull-upstream.sh
zijianjoy: The reason I ask is that you have already created a pull-upstream make target in https://github.com/kubeflow/gcp-blueprints/blob/2b8e52dd10796d78e40f40d6d9617b9e64b773b3/kubeflow/apps/pipelines/Makefile#L73, so I guess the goal is to use it. 

As mentioned in the other comment, if the eventual state is to separate pull-upstream commands to each sub-folder, or use kpt subpackage when it becomes available, I prefer to create a template script for pull-upstream, so each sub-folder can call this template from Makefile, instead of having pull-upstream.sh in each sub-folder. Because it will increase amount of files in the whole repo, and it is hard to keep them consistent. If there is no Makefile, then we will use the default template to pull upstream.

For this PR, feel free to keep the current usage. And we can update them at one go when we break down the kubeflow/pull-upstream.sh.

kubeflow/apps/pipelines/pull-upstream.sh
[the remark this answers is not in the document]
zijianjoy: Sounds good! Happy to try out sub package once available.

zijianjoy:
/lgtm

Waiting for upstream target to be changed to KFP 1.6.1.

kubeflow/pull-upstream.sh
Bobgy: Yes, I added the makefile rule for consistency.

> If there is no Makefile, then we will use the default template to pull upstream.

I think we need to know where upstream is and which version, these info need to be somewhere in the folder.

Bobgy: Logged in https://github.com/kubeflow/gcp-blueprints/issues/287

Ark-kun:
Perhaps we can rename this PR "make KFP v2 compatible mode work out of the box".

Bobgy:
/hold
still need to wait

Bobgy:
/unhold
Yes, this is ready for merge.

PR #285, opened 2021-06-01
[KF 1.3.1] Use kubeflow/manifests for pipelines upstream

zijianjoy:
As described in the title, we used to point pipeline upstream directly at kubeflow/pipelines. The common approach should be use kubeflow/manifests.

Consider picking it up in Kubeflow 1.3.1 release

### DIFF
--- a/kubeflow/pull-upstream.sh
+++ b/kubeflow/pull-upstream.sh
@@ -5,9 +5,6 @@ set -ex
 export KUBEFLOW_MANIFESTS_VERSION=v1.3.0
 export KUBEFLOW_MANIFESTS_REPO=https://github.com/kubeflow/manifests.git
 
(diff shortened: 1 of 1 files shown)


Bobgy:
I feel that it's better we use kubeflow/pipelines manifests directly, to make it easier for users to upgrade/downgrade pipelines.
kubeflow/manifests can just be the reference

Or we could use one as default, but comment another as an option

zijianjoy:
> I feel that it's better we use kubeflow/pipelines manifests directly, to make it easier for users to upgrade/downgrade pipelines.
> kubeflow/manifests can just be the reference
> 
> Or we could use one as default, but comment another as an option

Closing this PR since we are going to adopt KFP directly from KFP repo.

PR #286, opened 2021-06-01
[KF 1.3.1] Pin image type to be docker

zijianjoy:
Reference: https://github.com/kubeflow/pipelines/issues/5714

Make sure that Kubeflow cluster is using docker image type, otherwise pipeline will fail. 

Also related: https://github.com/kubeflow/gcp-blueprints/issues/271#issuecomment-846299401


Bobgy:
/lgtm

Thank you!

PR #314, opened 2021-09-19
kustomize: Use labelTransformer to replace commonLabels. Fix #298

### LINKED ISSUE #298: Consider removing commonLabels kustomize setter for `common` components
### Context
https://github.com/kubeflow/pipelines/issues/3755
https://github.com/kubeflow/gcp-blueprints/pull/212#issuecomment-823482800
https://github.com/kubeflow/kfserving/issues/1660#issuecomment-896390630

### Description

The `commonLabels` field in kustomization.yaml will apply labels to both the `labels` field and the `selectors` field of resource. The original description of this issue is in https://github.com/kubernetes-sigs/kustomize/issues/1009. I have fell into this trap for multiple times when maintaining GCP blueprint repo, and it took me long time trying to debug and identity root cause.

Because the components in `kubeflow/common` directory are not Kubeflow specific application, I don't think it applies to https://github.com/kubeflow/community/blob/master/guidelines/application_requirements.md. Therefore I tend to remove the `commonLabels` in their kustomization.yaml files. Example:
https://github.com/kubeflow/gcp-blueprints/blob/master/kubeflow/common/knative/kustomization.yaml#L7-L10

cc @Bobgy

zijianjoy:
Fix #298.

- Remove knative-serving namespace in kustomization, it updated knative-local-gateway to deploy in istio-system namespace.
- Use labelTransformer to allow knative-local-gateway to select istio-ingressgateway.

### DIFF
--- a/kubeflow/apps/profiles/kustomization.yaml
+++ b/kubeflow/apps/profiles/kustomization.yaml
@@ -16,8 +16,6 @@ resources:
 # - ../../../kubeflow/upstream/manifests/apps/profiles/upstream/overlays/istio/virtual-service.yaml
 # - ../../../kubeflow/upstream/manifests/apps/profiles/upstream/overlays/application/application.yaml
 
-commonLabels:
-  kustomize.component: profiles
 patchesStrategicMerge:
 - patches/manager.yaml
 - patches/profiles-config.yaml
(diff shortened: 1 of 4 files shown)


Bobgy (commented):
/lgtm

Issue #330, opened 2021-10-17
Issue with Deploy Kubeflow Cluster


akhilvinvent:
Hi,

I am facing issue with "Deploy Kubeflow Cluster" steps mentioned at below link.

Below upstream directory was not created after running "bash ./pull-upstream.sh" mentioned in document.

gcp-blueprints/kubeflow/apps/pipelines/upstream

https://www.kubeflow.org/docs/distributions/gke/deploy/deploy-cli/

Error: ./gcp-blueprints/kubeflow/apps/pipelines/upstream/third-party/argo/upstream/manifests/namespace-install/overlays/argo-server-deployment.yaml: wrong Node Kind for  expected: MappingNode was SequenceNode: value: {- op: add
  path: /spec/template/spec/containers/0/args/-
  value: --namespaced}

edi-bice-by:
kubeflow apply fails at the install_asm step for me

install_asm: Downloading kpt..
...
install_asm: Downloading ASM..
...
install_asm: Downloading ASM kpt package...
install_asm: [WARNING]: Failed, retrying...(1 of 3)
install_asm: [WARNING]: Failed, retrying...(2 of 3)
install_asm: [WARNING]: Failed, retrying...(3 of 3)
make: *** [install-asm] Error 1

Stepping through and trying to do manually using latest kpt

bash-5.1$ kpt pkg get https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages.git/asm@1.10.4-asm.6+config2 asm
Package "asm":
Fetching https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages@1.10.4-asm.6+config2
From https://github.com/GoogleCloudPlatform/anthos-service-mesh-packages
 * tag               1.10.4-asm.6+config2 -> FETCH_HEAD
Error: Kptfile at "https:/github.com/GoogleCloudPlatform/anthos-service-mesh-packages/asm@1.10.4-asm.6+config2" has an old version ("v1alpha1") of the Kptfile schema.
Please update the package to the latest format by following https://kpt.dev/installation/migration. 
bash-5.1$ kpt version
1.0.0-beta.7

edi-bice-by:
Not sure if it matters for kpt but Darwin was set to arm64 in the following code 

`download_kpt() {
  local OS

  case "$(uname)" in
    Linux ) OS="linux_amd64";;
    Darwin) OS="darwin_amd64";;
[...]

PR #351, opened 2022-02-22
Create dirs for kserve and models-web-app

LeonardAukea:
The script fails due to missing directories  `contrib/kserve/kserve/upstream` & `mkdir -p contrib/kserve/models-web-app/upstream`.

This PR makes sure to create them.

### DIFF
--- a/kubeflow/pull-upstream.sh
+++ b/kubeflow/pull-upstream.sh
@@ -158,6 +158,7 @@ rm contrib/metacontroller/upstream/Kptfile
 if [ -d contrib/kserve/models-web-app/upstream ]; then
     rm -rf contrib/kserve/models-web-app/upstream
 fi
+mkdir -p contrib/kserve/models-web-app/upstream
 kpt pkg get "${KUBEFLOW_MANIFESTS_REPO}/contrib/kserve/models-web-app/@${KUBEFLOW_MANIFESTS_VERSION}" contrib/kserve/models-web-app/upstream
 rm contrib/kserve/models-web-app/upstream/Kptfile
 
@@ -165,5 +166,6 @@ rm contrib/kserve/models-web-app/upstream/Kptfile
 if [ -d contrib/kserve/kserve/upstream ]; then
     rm -rf contrib/kserve/kserve/upstream
 fi
+mkdir -p contrib/kserve/kserve/upstream
 kpt pkg get "${KUBEFLOW_MANIFESTS_REPO}/contrib/kserve/kserve/@${KUBEFLOW_MANIFESTS_VERSION}" contrib/kserve/kserve/upstream
 rm contrib/kserve/kserve/upstream/Kptfile


zijianjoy:
Thank you @LeonardAukea , may I ask which OS platform you are running on and what is the kpt version you use? My script should automatically create a upstream folder like the rest of the script does.

LeonardAukea:
> Thank you @LeonardAukea , may I ask which OS platform you are running on and what is the kpt version you use? My script should automatically create a upstream folder like the rest of the script does.

OSX 12.2.1 and the latest version of `kpt`.  Why would you be expecting it to automatically create the dir in these two isolated cases? In the rest of the services you are explicitly creating the dir beforehand. Also it errored out for me when running it locally. 

Furthermore,  I think here the path was meant to be `apps/profiles/upstream` instead of `apps/profiles`

https://github.com/kubeflow/gcp-blueprints/blob/f4ff24a8acedb2d90d5a779e6fe5a338c71de345/kubeflow/pull-upstream.sh#L57-L58

zijianjoy:
@LeonardAukea My machine is Linux, so my guess is that kpt behavior is different between OSX and Linux. If I apply your fix, kpt will create a folder `models-web-app` under `upstream`. But if I don't use your fix, kpt will create a `upstream` folder and put manifests directly under it. 

In order to make both systems work, I have made the above suggestion, would you like to test it out on your side to see whether it works? If so, I will approve this PR. Thank you for your contribuiton!

kindomLee:
I think we still need to create this folder (`contrib/kserve/kserve`) to avoid this problems

```sh
Fetched 1 package(s).
+ rm contrib/kserve/models-web-app/upstream/Kptfile
+ '[' -d contrib/kserve/kserve/upstream ']'
+ kpt pkg get https://github.com/kubeflow/manifests.git/contrib/kserve/kserve/@v1.5.0 contrib/kserve/kserve/upstream
Error: parent directory "contrib/kserve/kserve" does not exist
```

zijianjoy:
@kindomLee Thank you for checking, I can confirm that your argument is correct. Unfortunately I run out of ideas about how to write this part of script to work on both Linux and MacOS. Do you have any suggestion? I can help validate your approach on Linux.

zijianjoy:
@kindomLee If you run this command, you should work on Mac:

```
kpt pkg get https://github.com/kubeflow/manifests.git/contrib/kserve/kserve@v1.5.0 contrib/kserve/kserve/kserve
```

I think it is kpt which doesn't have consistent behavior between Mac and Linux, I would suggest creating a new issue on kpt as well.

kindomLee:
In fact, I did not encounter this problem, I can create the upstream folder both in linux and OSX without any problem.
Only in the last encountered the situation that the folder `/contrib/kserve/kserve/` is not found and failed.
So whether it's in github or script, just adding this folder will solve my problem.

However, because of the #330 problem, I changed the installation process to be done by linux.

zijianjoy:
@kindomLee Note that the issue is not where the command failed, but in the `kpt` command.

Take a look at this command, the last argument is `contrib/kserve/kserve`:
```
kpt pkg get https://github.com/kubeflow/manifests.git/contrib/kserve/kserve/@v1.5.0 contrib/kserve/kserve/
```

On Linux:
Its behavior is to create a folder `contrib/kserve/kserve/kserve`, and put content inside this new folder.

On Mac:
Its behavior is to reuse existing folder `contrib/kserve/kserve`, and put content inside this existing folder.


As a result, you see error `'contrib/kserve/kserve/kserve': No such file or directory` on Mac, but not on Linux. Because the same kpt command on Linux created this new folder, but this folder is not created on Mac.

kindomLee:
@zijianjoy I don't think so, I get the same error results with the suggested change script on both OS

> On Linux
```sh
> ls -al
total 8
drwxrwxr-x  2 mark.lee mark.lee 4096 Mar 24 09:02 .
drwxr-xr-x 18 mark.lee mark.lee 4096 Mar 23 15:00 ..
> mkdir -p contrib/kserve
> kpt pkg get https://github.com/kubeflow/manifests.git/contrib/kserve/kserve/@v1.5.0 contrib/kserve/kserve/
Package "kserve":
Fetching https://github.com/kubeflow/manifests@v1.5.0
From https://github.com/kubeflow/manifests
 * tag               v1.5.0     -> FETCH_HEAD
Adding package "contrib/kserve/kserve".

Fetched 1 package(s).
> mv contrib/kserve/kserve/kserve contrib/kserve/kserve/upstream
mv: cannot stat 'contrib/kserve/kserve/kserve': No such file or directory
> ls contrib/kserve/kserve/
Kptfile  aggregated-roles.yaml  kserve.yaml  kserve_kubeflow.yaml  kustomization.yaml
> echo "$SHELL"
/bin/bash
```

> On Mac
```sh
> ls -al
total 0
drwxr-xr-x@  2 mark.lee  staff   64  3 24 09:06 .
drwxr-xr-x@ 28 mark.lee  staff  896  2 25 12:03 ..
> mkdir -p contrib/kserve
> kpt pkg get https://github.com/kubeflow/manifests.git/contrib/kserve/kserve/@v1.5.0 contrib/kserve/kserve/
Package "kserve":
Fetching https://github.com/kubeflow/manifests@v1.5.0
From https://github.com/kubeflow/manifests
 * tag               v1.5.0     -> FETCH_HEAD
Adding package "contrib/kserve/kserve".

Fetched 1 package(s).
> mv contrib/kserve/kserve/kserve contrib/kserve/kserve/upstream
mv: rename contrib/kserve/kserve/kserve to contrib/kserve/kserve/upstream: No such file or directory
> ls contrib/kserve/kserve
Kptfile               aggregated-roles.yaml kserve.yaml           kserve_kubeflow.yaml  kustomization.yaml
> echo "$SHELL"
/bin/zsh
```

zijianjoy:
@kindomLee Good to know! I am guessing I might be using an old version of kpt then. Because this behavior is different from what I experience. Would you like to tell me what is your kpt version on both Linux and Mac?

kindomLee:
> Linux
`1.0.0-beta.13`

> OSX
`1.0.0-beta.7`

LoL

zijianjoy:
Okay I have upgraded kpt to the versions you have, I have validated the following command to be working on both OS, do you want to help checking it works on your side? @kindomLee 

```

if [ -d contrib/kserve/kserve/upstream ]; then
    rm -rf contrib/kserve/kserve/upstream
fi
kpt pkg get "${KUBEFLOW_MANIFESTS_REPO}/contrib/kserve/kserve/@${KUBEFLOW_MANIFESTS_VERSION}" contrib/kserve/kserve-upstream/
mkdir -p contrib/kserve/kserve/upstream
mv contrib/kserve/kserve-upstream/* contrib/kserve/kserve/upstream/
rmdir contrib/kserve/kserve-upstream
rm contrib/kserve/kserve/upstream/Kptfile
```

kindomLee:
This works for me, but how about this one?

```
if [ -d contrib/kserve/kserve/upstream ]; then
    rm -rf contrib/kserve/kserve/upstream
fi
mkdir -p contrib/kserve/kserve
kpt pkg get "${KUBEFLOW_MANIFESTS_REPO}/contrib/kserve/kserve/@${KUBEFLOW_MANIFESTS_VERSION}" contrib/kserve/kserve/upstream/
rm contrib/kserve/kserve/upstream/Kptfile
```

zijianjoy:
@kindomLee Very good suggestion! I like the approach you proposed, and it works on my Linux machine. Do you want to create a PR using this approach?

LeonardAukea:
I switched back to old school bash scripts as `kpt` was misbehaving even on the same OS. Unfortunate, it seems like a cool tool.

kindomLee:
@LeonardAukea I believe it will be more stable when they remove the `beta` suffix.

leetdavid:
Will this be merged anytime soon? I'm on Ubuntu 20.04 on WSL on Windows 10 (and my coworker is on macOS) and we currently still need to create the kserve folder manually.

zijianjoy:
@leetdavid Does this merged PR help you and your co-worker? https://github.com/kubeflow/gcp-blueprints/pull/361

Arhell (commented):
/ok-to-test

zijianjoy:
Closing this PR as we have already implemented the fix in master branch. Feel free to comment if you need further help on it.

PR #358, opened 2022-03-22
Fix KF_NAME length

kindomLee:
* update env.sh
> Conflict with iam.gserviceaccount.com limit length(6~30)

### DIFF
--- a/kubeflow/apps/pipelines/pull-upstream.sh
+++ b/kubeflow/apps/pipelines/pull-upstream.sh
@@ -17,7 +17,7 @@
 set -ex
 
 # TODO: Use kubeflow/pipelines once https://github.com/kubeflow/pipelines/pull/6595 is resolved.
-export KUBEFLOW_PIPELINES_VERSION=1.8.1-rc.0
+export KUBEFLOW_PIPELINES_VERSION=1.8.1
 export KUBEFLOW_PIPELINES_REPO=https://github.com/kubeflow/pipelines.git
 # export KUBEFLOW_PIPELINES_VERSION=upgradekpt # Other attempted branches: krmignore, kubeflow14
 # export KUBEFLOW_PIPELINES_REPO=https://github.com/zijianjoy/pipelines.git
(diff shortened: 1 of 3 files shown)


kindomLee:
[Affected files](kubeflow/apps/pipelines/cloudsql/proxy/workload-identity-binding-patch.yaml)

```sh
# kpt-set: ${name}-kfp-cloudsql@${gcloud.core.project}.iam.gserviceaccount.com
```

Arhell (commented):
Please sign CLA
https://github.com/kubeflow/gcp-blueprints/pull/358/checks?check_run_id=5640630849

kindomLee:
> Please sign CLA https://github.com/kubeflow/gcp-blueprints/pull/358/checks?check_run_id=5640630849

I have signed in both of my accounts, but it doesn't seem to be rescan successful?

image

zijianjoy:
Thank you for the PR! Indeed it is a problem where we should fix the name length restriction.

Instead of shortening from 24 to 17, how do you think about making the cloudsql service account shorter, so we give people more characters to set their project? For example: <kubeflow-name>-sql@iam.gserviceaccount.com.

You can check the CLA issue here, looks like it is related to email: https://github.com/kubeflow/gcp-blueprints/pull/358/checks?check_run_id=5653631084

kindomLee:
I think it's a good idea, so we don't have to make changes in `env.sh` and less impact on users

And I screwed up commit author information...

> Thank you for the PR! Indeed it is a problem where we should fix the name length restriction.
> 
> Instead of shortening from 24 to 17, how do you think about making the cloudsql service account shorter, so we give people more characters to set their project? For example: `<kubeflow-name>-sql@iam.gserviceaccount.com`.
> 
> You can check the CLA issue here, looks like it is related to email: https://github.com/kubeflow/gcp-blueprints/pull/358/checks?check_run_id=5653631084

zijianjoy:
@kindomLee Thank you for the confirmation! I can review the PR once you have updated it to make the service account name suffix shorter. 

You can configure your email by using:
```
git config --global user.email "your_email@abc.com"
```
Then amend your existing commit

```
git commit --amend --reset-author
```

Then force push to your remote branch to overwrite your existing commit, in order to be CLA compliance.

PR #359, opened 2022-03-23
cloudsql: Use a shorter cloudsql service account name to allow longer Kubeflow name

kindomLee:
fix the name length restriction #358

### DIFF
--- a/kubeflow/apps/pipelines/cloudsql/cnrm/iam.yaml
+++ b/kubeflow/apps/pipelines/cloudsql/cnrm/iam.yaml
@@ -5,7 +5,7 @@ metadata:
 spec:
   memberFrom:
     serviceAccountRef:
-      name: KUBEFLOW-NAME-kfp-cloudsql # kpt-set: ${name}-kfp-cloudsql
+      name: KUBEFLOW-NAME-sql # kpt-set: ${name}-sql
   resourceRef:
     apiVersion: resourcemanager.cnrm.cloud.google.com/v1beta1
     external: projects/PROJECT # kpt-set: projects/${gcloud.core.project}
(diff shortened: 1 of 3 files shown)


zijianjoy:
Have you validated this change works?

We use this service account in the following files:

- https://github.com/kubeflow/gcp-blueprints/blob/d96b7fe4a95d306acc36387151947b27ea023a96/kubeflow/apps/pipelines/cloudsql/cnrm/iam.yaml
- https://github.com/kubeflow/gcp-blueprints/blob/d96b7fe4a95d306acc36387151947b27ea023a96/kubeflow/apps/pipelines/cloudsql/cnrm/proxy-gsa.yaml

Note: Even with the complete change, I think it will still cause potential breaking change for Kubeflow upgrade scenario. But let's focus on the first time installation scenario for now.

kindomLee:
You are right, I will test the complete installation process in my environment and modify the files that are affected.

kindomLee:
> Have you validated this change works?
> 
> We use this service account in the following files:
> 
> * https://github.com/kubeflow/gcp-blueprints/blob/d96b7fe4a95d306acc36387151947b27ea023a96/kubeflow/apps/pipelines/cloudsql/cnrm/iam.yaml
> * https://github.com/kubeflow/gcp-blueprints/blob/d96b7fe4a95d306acc36387151947b27ea023a96/kubeflow/apps/pipelines/cloudsql/cnrm/proxy-gsa.yaml
> 
> Note: Even with the complete change, I think it will still cause potential breaking change for Kubeflow upgrade scenario. But let's focus on the first time installation scenario for now.

I completed the installation process once using the default values and successfully logged in via google iap authentication

zijianjoy:
/lgtm

Thank you for validating! But I think login is not enough to validate the change working. Because you need to use CloudSQL by running a Kubeflow Pipeline. 
LGTM on the change itself, pending final validation to provide approval.

kindomLee:
image
I think we are good to go

zijianjoy:
/approve

Thank you so much for the contribution!

PR #361, opened 2022-03-24
fix(pull-upstream): create kserve directory to store upstream files

kindomLee:
create kserve directory to store upstream files and prevent `no such file or directory` error


zijianjoy:
/lgtm
/approve

PR #364, opened 2022-04-12
Move `RequestAuthentication` policy creation to `iap-enabler` for gitops friendliness

fabito:
# Problem

In my gitops setup I was facing a race condition between `Fluxcd` and `iap-enabler` where they were both updating the `ingress-jwt` RequestAuthentication resource:

- Fluxcd creates/updates the `ingress-jwt` RequestAuthentication resource without the audience (TO_BE_PATCHED)
- `iap-enabler` resolves the backend id and patches the `ingress-jwt` audience

# Solution

Move `ingress-jwt` RequestAuthentication creation logic to `iap-enabler`.

- Move `policy.yaml` from Kustomize resources list to `envoy-config` ConfigMap
- Refactor `setup_backend.sh` to replace `TO_BE_PATCHED` with the correct audience and apply `policy.yaml` to create/update `ingress-jwt`

### DIFF
--- a/kubeflow/common/iap-ingress/base/policy.yaml
+++ b/kubeflow/common/iap-ingress/base/policy.yaml
@@ -8,7 +8,7 @@ spec:
       app: istio-ingressgateway
   jwtRules:
   - audiences:
-    - TO_BE_PATCHED
+    - JWT_AUDIENCE
     issuer: https://cloud.google.com/iap
     jwksUri: https://www.gstatic.com/iap/verify/public_key-jwk
     fromHeaders:
--- a/kubeflow/common/iap-ingress/base/config-map.yaml
+++ b/kubeflow/common/iap-ingress/base/config-map.yaml
@@ -55,6 +55,8 @@ data:
     [ -z ${SERVICE} ] && echo Error SERVICE must be set && exit 1
     [ -z ${INGRESS_NAME} ] && echo Error INGRESS_NAME must be set && exit 1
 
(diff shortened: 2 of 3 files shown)


Arhell (commented):
/ok-to-test

zijianjoy:
Thank you @fabito , will this change make `policy.yaml` out of the control from gitops `Fluxcd`? Also, should we use more specific placeholder than `TO_BE_PATCHED` since we use `sed` to replace the value now? Previously we specify the yaml path for the `JWT_AUDIENCE` replacement.

fabito:
> Thank you @fabito , will this change make `policy.yaml` out of the control from gitops `Fluxcd`? 
Yep, solely managed by `iap-enabler` now

> Also, should we use more specific placeholder than `TO_BE_PATCHED` since we use `sed` to replace the value now? 
Sure!

kubeflow/common/iap-ingress/base/policy.yaml
```diff
@@ -8,7 +8,7 @@ spec:
       app: istio-ingressgateway
   jwtRules:
   - audiences:
-    - TO_BE_PATCHED
+    - JWT_AUDIENCE
```
fabito, line 11: @zijianjoy , let me know if you have a better name in mind.

zijianjoy: This name sounds good to me, thanks!

PR #397, opened 2022-11-17
Add script to automatically create OAuth clients on GCP

richardsliu:
Using documentation here: https://cloud.google.com/iap/docs/programmatic-oauth-clients

Script will output CLIENT_ID and CLIENT_SECRET values.

This is to replace manual setup steps described in https://www.kubeflow.org/docs/distributions/gke/deploy/oauth-setup/.

### DIFF
--- a/kubeflow/hack/setup_oauth_client.sh
+++ b/kubeflow/hack/setup_oauth_client.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+#
+# A helper script to create OAuth clients using the caller's GCP project.
+#
+
+if [[ ! -n $ADMIN_EMAIL ]]; then
+  echo "Please provide a value for ADMIN_EMAIL"
+  exit 1
+fi
+
+if [[ ! -n $NAME ]]; then
+  echo "Please provide a value for NAME"
+  exit 1
+fi
+
+BRAND_ID="`gcloud iap oauth-brands list | grep 'name: '`"
+
+if [[ ! -n $BRAND_ID ]]; then
+  echo 'OAuth brand does not exist, creating new one...'
+  BRAND_ID="`gcloud iap oauth-brands create --application_title=kubeflow --support_email=$ADMIN_EMAIL | grep 'name: '`"
+fi
+
+BRAND_ID=$(echo $BRAND_ID | sed -e "s/name: //g")
+
+CLIENT_OUTPUT=`gcloud iap oauth-clients create $BRAND_ID --display_name=$NAME`
+
+export CLIENT_ID=$(echo $CLIENT_OUTPUT | sed 's/.*identityAwareProxyClients\/\([^ ]*\).*/\1/')
+echo "CLIENT_ID=$CLIENT_ID"
+
+export CLIENT_SECRET=$(echo $CLIENT_OUTPUT | sed 's/.*secret: \([^ ]*\).*/\1/')
+echo "CLIENT_SECRET=$CLIENT_SECRET"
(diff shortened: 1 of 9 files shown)


kubeflow/hack/setup_oauth_client.sh
```diff
@@ -0,0 +1,31 @@
...
+fi
+
+BRAND_ID="`gcloud iap oauth-brands list | grep 'name: '`"
+
+if [[ ! -n $BRAND_ID ]]; then
+  echo 'OAuth brand does not exist, creating new one...'
+  BRAND_ID="`gcloud iap oauth-brands create --application_title=kubeflow --support_email=$ADMIN_EMAIL | grep 'name: '`"
+fi
+
+BRAND_ID=$(echo $BRAND_ID | sed -e "s/name: //g")
```
zijianjoy, line 23: Can we remove the back quote symbol '`' from $BRAND_ID assignment, and use `eval` instead of `echo` here?

kubeflow/hack/setup_oauth_client.sh
```diff
@@ -0,0 +1,31 @@
...
+
+CLIENT_OUTPUT=`gcloud iap oauth-clients create $BRAND_ID --display_name=$NAME`
+
+export CLIENT_ID=$(echo $CLIENT_OUTPUT | sed 's/.*identityAwareProxyClients\/\([^ ]*\).*/\1/')
+echo "CLIENT_ID=$CLIENT_ID"
+
+export CLIENT_SECRET=$(echo $CLIENT_OUTPUT | sed 's/.*secret: \([^ ]*\).*/\1/')
```
gkcalat, line 30: If we quietly exporting these variable, should we `unset BRAND_ID CLIENT_ID CLIENT_SECRET CLIENT_OUTPUT` somewhere after deploying to prevent secrets from being kept in environmental variables?

kubeflow/hack/setup_oauth_client.sh
```diff
@@ -0,0 +1,31 @@
...
+BRAND_ID="`gcloud iap oauth-brands list | grep 'name: '`"
+
+if [[ ! -n $BRAND_ID ]]; then
+  echo 'OAuth brand does not exist, creating new one...'
+  BRAND_ID="`gcloud iap oauth-brands create --application_title=kubeflow --support_email=$ADMIN_EMAIL | grep 'name: '`"
+fi
+
+BRAND_ID=$(echo $BRAND_ID | sed -e "s/name: //g")
+
+CLIENT_OUTPUT=`gcloud iap oauth-clients create $BRAND_ID --display_name=$NAME`
```
gkcalat, line 25: This will always create a new OAuth client ID. If this is run multiple times, multiple client ids will be create with the same client name. Should we first use `list` to check for existing ids?

kubeflow/hack/setup_oauth_client.sh
```diff
@@ -0,0 +1,31 @@
...
+#
+# A helper script to create OAuth clients using the caller's GCP project.
+#
+
+if [[ ! -n $ADMIN_EMAIL ]]; then
+  echo "Please provide a value for ADMIN_EMAIL"
+  exit 1
+fi
+
+if [[ ! -n $NAME ]]; then
```
gkcalat, line 11: We use `KF_NAME` as a name for the Kubeflow cluster. Can we rename this variable into `OAUTH_CLIENT_NAME`?

kubeflow/hack/setup_oauth_client.sh
richardsliu: Perhaps it is better to not set these at all, but parse the results in the IAP setup Makefile instead?

kubeflow/hack/setup_oauth_client.sh
richardsliu: The OAuth client name is not really significant and one seems sufficient for a domain, should we just use some default name here?

PR #398, opened 2022-11-21
Set node image type to cos_containerd

thomas-vl:
Because docker based images are no longer supported in the regular channel we need to specify a containerd image for the node pool


zijianjoy:
Thank you @thomas-vl ! Since the default image type on GKE is cos_containerd. I think we can close this PR now.

PR #399, opened 2022-12-02
Create a script to programatically create OAuth secret for IAP

richardsliu:
This is a continuation of the work in https://github.com/GoogleCloudPlatform/kubeflow-distribution/pull/397.

This change integrates the `setup_iap_ingress.sh` script into the Makefile so the OAuth client and secret are automatically created if they do not exist. If the `CLIENT_ID` and `CLIENT_SECRET` environment variables are created then the script will simply use them to create the IAP secret.

### DIFF
--- a/kubeflow/hack/setup_iap_ingress.sh
+++ b/kubeflow/hack/setup_iap_ingress.sh
@@ -1,0 +1,48 @@
+#!/bin/bash
+#
+# A helper script to create IAP ingress using the caller's GCP project.
+# The script will first create OAuth client and brand if these are
+# not provided.
+#
+# To find more information about the commands, please see
+# https://cloud.google.com/iap/docs/programmatic-oauth-clients.
+#
+#
+if [[ ! -n $KFCTXT ]]; then
+  echo "Please provide a value for KFCTXT"
+  exit 1
+fi
+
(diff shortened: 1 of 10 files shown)


kubeflow/hack/setup_iap_ingress.sh
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash
+#
+# A helper script to create IAP ingress using the caller's GCP project.
+# The script will first create OAuth client and brand if these are
+# not provided.
+#
+# To find more information about the commands, please see
+# https://cloud.google.com/iap/docs/programmatic-oauth-clients.
```
zijianjoy, line 8: How do we provide redirect URI via command line approach? https://www.kubeflow.org/docs/distributions/gke/deploy/oauth-setup/

richardsliu: Seems like redirect URIs are not supported in the commandline. Should we provide an option that skips this step if the user needs redirects?

zijianjoy: I think redirects is required for Kubeflow deployment to work, please correct me if I am wrong.

cc @gkcalat

PR #409, opened 2023-03-12
Upgrade cert-manager to 1.10.2. Closes #405.

### LINKED ISSUE #405: Kubeflow 1.7: Update cert-manager to 1.10.1
cert-manager `1.5.0` -> `1.10.1`

Reference: [Kubeflow 1.7 tracker](https://github.com/kubeflow/manifests/issues/2359)

gkcalat:
Previous upgrade: [cert-manger 1.5.0](https://github.com/GoogleCloudPlatform/kubeflow-distribution/pull/372)

- Updated the upstream version
- Changed the folder name
- Updated .krmignore
- [x] Tested on GKE 1.23

Full testing on GKE 1.24 or 1.25 will be done once all components will be upgraded.

I chose `1.10.2` over `1.10.1` as the latest patch [included](https://cert-manager.io/docs/release-notes/release-notes-1.10#v1102-changes-since-v1101) vulnerability fixes and a significant performance boost.


zijianjoy (commented):
It looks good to me, but do we want to remove version number `cert-manager-1-10` from the folder name? In this case, we just need to pull from upstream and rely on it for cert manager version.

zijianjoy (approved):

gkcalat:
> It looks good to me, but do we want to remove version number `cert-manager-1-10` from the folder name? In this case, we just need to pull from upstream and rely on it for cert manager version.

Thank you! Updated.

kubeflow/common/cert-manager/kustomization.yaml
```diff
@@ -1,4 +1,4 @@
 apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 resources:
-- ./cert-manager-1-10/cert-manager/base
+- ./cert-manager/base
```
zijianjoy, line 4: Shouldn't it be `cert-manager/cert-manager/base` instead?

gkcalat: I removed the `cert-manager-1-10`, which, besides removing the version from the folder name, also removed the unnecessary parent folder. This `kustomization.yaml` compiles and cert-manager deploys successfully.

zijianjoy: That makes sense, thank you Ablai!

zijianjoy (approved):
Thank you Ablai for the contribution!

Issue #414, opened 2023-03-17
[AutoML] Missing user header


gkcalat:
`katib` has a breaking change in `0.15-rc.1`, which is being used in Kubeflow `1.7`.

Detail: https://github.com/kubeflow/katib/issues/2126.

PR #415, opened 2023-03-17
fix(katib): Add user header to address #414

gkcalat:
- Added a patch to `katib-ui` deployment
- [x] Verified on GKE `1.23` with `1.7.0-rc.2` manifests

Reference: https://github.com/kubeflow/katib/issues/2126


zijianjoy:
Great job, Ablai! Your work looks good to me.

PR #417, opened 2023-03-23
Upgrade KServe, upstream manifests

gkcalat:
Changes:
- [x] Upgraded `KServe` to `0.10` ([release notes with breaking changes](https://github.com/kserve/kserve/releases))
   - [x] Removed deprecated `kserve-config` ConfigMap
   - [x] Replaced StatefulSet with Deployment for `kserve-controller-manager`
   - [x] Removed obsolete patches
- [x] Upgraded upstream manifests to `1.7.0-rc.2`
- [x] Upgraded KFP to `2.0.0-alpha.7`
- [x] Tested n GKE `1.23` and GKE `1.24`


kubeflow/contrib/kserve/patches/namespace.yaml
```diff
@@ -1,6 +0,0 @@
-# Remove namespace resource as namespace will already exist.
-$patch: delete
-apiVersion: v1
-kind: Namespace
-metadata: # kpt-merge: /kserve
-  name: kserve
```
zijianjoy, line 6: We should remove kserve namespace, right? Or is upstream handling this case?

kubeflow/contrib/kserve/patches/config-param.yaml
```diff
@@ -1,4 +0,0 @@
-- op: replace
-  path: /data
-  value:
-    ingressGateway: kubeflow/kubeflow-gateway
```
zijianjoy, line 4: How is ingressGateway being set in this upgrade?

kubeflow/contrib/kserve/patches/namespace.yaml
gkcalat: I don't see this namespace. The patch is not used anywhere either. We can keep it if you have a strong opinion.

I could find the namespace after deployment:


**this commit**:
```bash
> kubectl get namespace

NAME              STATUS   AGE
ablai             Active   19h
asm-system        Active   23h
cert-manager      Active   23h
default           Active   23h
istio-system      Active   23h
knative-serving   Active   23h
kserve-test       Active   18h
kube-node-lease   Active   23h
kube-public       Active   23h
kube-system       Active   23h
kubeflow          Active   23h
```

**1.6.1**:
```bash
> kubectl get namespace

NAME              STATUS   AGE
ablai             Active   8d
asm-system        Active   9d
cert-manager      Active   9d
default           Active   9d
istio-system      Active   9d
knative-serving   Active   9d
kserve-test       Active   24h
kube-node-lease   Active   9d
kube-public       Active   9d
kube-system       Active   9d
kubeflow          Active   9d
```

The `kserve-test` and `ablai` namespaces were created during [kserve testing](https://kserve.github.io/website/master/get_started/first_isvc/#2-create-an-inferenceservice) and [e2d MNIST testing](https://github.com/kubeflow/pipelines/blob/master/samples/contrib/kubeflow-e2e-mnist/kubeflow-e2e-mnist.ipynb).

kubeflow/contrib/kserve/patches/config-param.yaml
gkcalat: This ConfigMap was removed as a part of cleanup in KServe 0.10. I believe this is now hardcoded in the `kubeflow_kserve.yaml`. Let me know what's your opinion on this. Thank you!

kubeflow/contrib/kserve/patches/namespace.yaml
zijianjoy: I think the intention of this file is to remove any pre-existing kserve deployment. Multiple kserve deployment in a single GKE cluster will likely cause problem.

kubeflow/contrib/kserve/patches/config-param.yaml
zijianjoy: Which file is kubeflow_kserve.yaml? Can you point me to the file? Thank you!

zijianjoy: I found it in https://github.com/kubeflow/manifests/blob/master/contrib/kserve/kserve/kserve_kubeflow.yaml, sounds good on removing this file then. Thank you!

kubeflow/contrib/kserve/patches/namespace.yaml
gkcalat: That makes sense for cases whith custom `kserve` installation. However, this patch has been commented out more [than year ago](https://github.com/GoogleCloudPlatform/kubeflow-distribution/pull/355) probably because it causes this error:
```
Error: no matches for Id ~G_v1_Namespace|~X|kserve; failed to find unique target for patch ~G_v1_Namespace|kserve
make: *** [Makefile:23: hydrate] Error 1
```

Should we keep it commented out and store the patch yaml (not sure is anyone will be affected if we remove the dead code)?

PR #441, opened 2023-09-28
auto remove lines with `creationTimestamp: "null"`

### LINKED ISSUE #440: kustomize generates file with creationTimestamp line
Similar to https://github.com/GoogleCloudPlatform/kubeflow-distribution/issues/431

Now Kustomize adds this line when [auto generating the kustomization files](https://github.com/GoogleCloudPlatform/kubeflow-distribution/blob/4945624167b18cc925fa0e8f4914d0d5e85631e6/kubeflow/hack/apply_component.sh#L37):

```
creationTimestamp: null
```

And it is not recognized by kubectl.  Since this issue emerged recently, it's likely caused by newer versions of kustomization using newer controller-gen.

Reference:
https://github.com/kubernetes-sigs/kustomize/issues/5031

Linchin:
fixes #440 

Adds a line to remove generated lines containing `creationTimestamp: "null"`, during kubectl having trouble reading them.


gkcalat (approved):
Thank you, Lingqing!

